PR #2, opened 2018-01-14
I make a simple module for polybar: system-nvidia-bbswitch

SeraphyBR:
I made this simple module for the polybar, to show if the nvidia graphics card are actually working in notebooks. It is possible that other people already have something of the type, but I decided to add this one to the repository.


x70b1:
This is awesome! Thanks!

PR #7, opened 2018-01-18
Add poll-less MPRIS music player script

alexozer:
It reacts to changes in track and play/pause status by listening to MPRIS events.


x70b1:
Thanks!

PR #8, opened 2018-01-20
Merge openweathermap scripts

alexozer:
Maybe this is better than maintaining four different weather scripts?


x70b1:
Thanks for the pull request. I'm not happy with the idea of merging the scripts at this point.

The goal should be to collect scripts so that they are customizable for everyone. We can't write scripts for every case. That's why I don't want an all-in-one script. The scripts should be easy to customize instead of covering every possible case.

Everyone wants something different. So he gets ideas and approaches here.

The Scripts `inbox-imap-*` and `inbox-pop3-*` are a very similar case. I could merge them. But I don't want to. Everyone can pick up what they need.

You forgot to delete your API key in the script. You should replace him now. :sunglasses:

alexozer:
I understand, I had a feeling it might be the case. And whoops, I tried
really hard to remember to do that, haha.

PR #9, opened 2018-01-20
Replace mpris-player with poll-less version

alexozer:
Might as well, right?


x70b1:
I've been thinking about it, too. But I decided not to. 

The same you can see at the `pulseaudio-*` scripts. There are two versions. A simple script and a version based on Polybars `tail` function. 

The point is, why throw away scripts that work and are up to date? I don't see any reason.

alexozer:
I understand. I guess my reasoning was that the "tail" versions should
somehow be considered "default", and that the simple scripts are only
really desirable if you're against a heavier Python runtime or something. I
guess people can be expected to choose the right version anyway though.

x70b1:
Yes, I also think that the tail versions are always preferred. But the `README.md` files are comprehensive, I think. Moreover, I believe that in case of doubt both scripts will be tested anyway. 

You PR contains a Commit `Parameterize icon for when no song playing`. Would you like to send a PR for that ? I would merge that commit.

alexozer:
Okay.

PR #22, opened 2018-01-26
Fixed cut error in info-redshift-temp.sh

srydell:
This script previously fetched ':' and tried to compare it to an integer. Changed it so that it fetches for example '5000K' and then cuts 'K' instead of the ':' from '..temperature: 5000K'.

Changed the command from
$ redshift -p 2> /dev/null | grep temp | cut -d " " -f 2 | tr -d "[:alpha:]"
to
$ redshift -p 2> /dev/null | grep temp | cut -d " " -f 3 | tr -d "[:alpha:]"

Explanation:
$ redshift -p 2> /dev/null | grep temp
yields
Color temperature: 5000K
piped into 'cut -d " " -f 2' gives
:
while piped into 'cut -d " " -f 3' gives
5000K
which is what we want.

Hope this helps!

### DIFF
--- a/polybar-scripts/info-redshift-temp/info-redshift-temp.sh
+++ b/polybar-scripts/info-redshift-temp/info-redshift-temp.sh
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 if [ "$(pgrep -x redshift)" ]; then
-    temp=$(redshift -p 2> /dev/null | grep temp | cut -d " " -f 3 | tr -d "[:alpha:]")
+    temp=$(redshift -p 2> /dev/null | grep temp | cut -d ":" -f 2 | tr -dc "[:digit:]")
 
     if [ -z "$temp" ]; then
         echo "%{F#65737E} #"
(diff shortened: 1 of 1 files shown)


x70b1:
Thank you for that fix. 
The script worked for me. But I will check it and merge your PR it if it is neccessary.

srydell:
Thank you for the great repository 👍

x70b1:
Okay, I found out that this is a problem when you use other translations. In german the output is `Farbtemperatur: 3500K` vs `Color temperature: 5000K` in english. That means if you cut the string at a space character you got different results.

Could you check the following?

```
temp=$(redshift -p 2> /dev/null | grep temp | cut -d ":" -f 2 | tr -dc "[:digit:]")
```

srydell:
I see. Good catch. Your code works for me.

```
$ temp=$(redshift -p 2> /dev/null | grep temp | cut -d ":" -f 2 | tr -dc "[:digit:]")
$ echo $temp
6500
```

I'll incorporate it into a commit as soon as I can.

x70b1:
Done :open_hands:

PR #26, opened 2018-02-04
Reset font after showing icon

132ikl:


x70b1:
Hi
Thanks for your pull request. 

Why do you want to reset the font?

132ikl:
The Weather-Icons font has a set of number built in which look like a pixel-font. This clashes with the rest of a polybar if you don't have yours set to a pixel font. This makes the text after the icon look normal again.

x70b1:
I can't reproduce that. 
What is the order of your font?

```ini
font-0 = ...
font-1 = ...
font-2 = Weather Icons:size=12;1
```

132ikl:
font-0 = fixed:pixelsize=10;1
font-1 = unifont:fontformat=truetype:size=8:antialias=false;0
font-2 = siji:pixelsize=10;1
font-3 = FontAwesome:size=12;0
font-4 = "Ubuntu Nerd Font:size=14;0"
font-5 = "Ubuntu Nerd Font:size=10;0"
font-6 = "Weather Icons:size=14;1"

On closer inspection, for the pixelsize font is font-0, but despite this my default font seems to be font-1. It's not being set for every module despite every other module using font-1, and the modification I made sets the text on the weather module to be the unifont.

x70b1:
It looks like an overlap between Weather Icons and FontAwesome. If that happens you have to reset the fonts. But I am not sure if this is an regular case.

PR #28, opened 2018-02-09
fix for issue : arithmetic expression: expecting primary

mackristof:
openweathermap-fullfeatured.sh: arithmetic expression: expecting primary: ""1518196038"-"1518189984""


x70b1:
You are right. It seems that the `dash` shell has problems at this point.

PR #34, opened 2018-03-06
Added pulseaudio-multiple-tail

bbaserdem:
Added a script that is a modification of pulseaudio-tail script, that allows for multiple sink info to be displayed.

### DIFF
--- a/polybar-scripts/pulseaudio-multiple-tail/README.md
+++ b/polybar-scripts/pulseaudio-multiple-tail/README.md
@@ -1,0 +1,37 @@
+# Script: { Pulseaudio-multi-tail }
+
+This script is a modification on the pulseaudio-tail script.
+
+* Adds icon detection for headphones or speakers on the analog sink
+* If you have either HDMI or Bluetooth speakers, detects them and displays their info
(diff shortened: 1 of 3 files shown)


x70b1:
Do you know already the new [`pulseaudio`](https://github.com/jaagr/polybar/wiki/Module:-pulseaudio) module? You can define a specific sink there.

Since there is now a real module for pulse, the shell scripts are a bit useless.

bbaserdem:
I did not see the ability auto detect weather the analog output is routed to speakers or headphones (laptop), and change the icon accordingly. Otherwise they have similar functionality

x70b1:
I just want to make sure we're not trying to solve a problem that's already solved. :v:

It looks like a script that shows the output sink would be good? For example:
* Icon 1 = headphones
* Icon 2 = speaker

Then we could combine it with the `pulseaudio` module on the bar.
What do you think?

x70b1:
@bbaserdem 
You can now define icons for different ports in the script.

x70b1:
There is already the same feature in `polybar/pulseaudio`:
https://github.com/jaagr/polybar/pull/1075

PR #38, opened 2018-04-05
Rewrite info-hackspeed

M1dgard:
Filter by key code to only count letters and numbers, and not interpunction and e.g. the Ctrl and Shift keys. (The way this works isn't foolproof but it should be good enough.)

Be more efficient: use less commands to achieve the same goal.

Add options, and make them configurable with environment variables. This would enable system-wide installations with per-user configurations. If you think the code gets too unorderly with the environment variable system, feel free to only merge the first and third commit and ignore the second.


x70b1:
Let me check it out.

x70b1:
Which icon font is that on your screenshot? It looks good!

x70b1:
You use `subtitles` from the `AV` section. :wink:

PR #50, opened 2018-05-31
reddit-inbox: Add option to show even if empty

132ikl:


132ikl:
wait i'm stupid

x70b1:
Yes, just print the var in every case :wink:

PR #51, opened 2018-06-03
Fix openweathermap scripts for when the API returns more than icon

davialexandre:
According to the [OpenWeatherMap API documentation](https://openweathermap.org/weather-data), it is possible for it to return more than one object in the `weather` array. Example:

```json
...
"weather":[
                 {"id":520,"main":"rain","description":"light intensity shower rain","icon":"09d"},
                 {"id":500,"main":"rain","description":"light rain","icon":"10d"},
                 {"id":701,"main":"mist","description":"mist","icon":"50d"}
              ],
...
```

The jq query used to get the icon from the API response would return a list of all the icons in each of the weather objects and pass it to the `get_icon` function. The function would not identify it as a valid value and return the default one instead.

To fix this, I changed the query to only return the icon of the first object in the list.

### DIFF
--- a/polybar-scripts/openweathermap-detailed/openweathermap-detailed.sh
+++ b/polybar-scripts/openweathermap-detailed/openweathermap-detailed.sh
@@ -44,9 +44,9 @@ else
 fi
 
 if [ ! -z "$weather" ]; then
-    weather_desc=$(echo "$weather" | jq -r ".weather[].description")
+    weather_desc=$(echo "$weather" | jq -r ".weather[0].description")
     weather_temp=$(echo "$weather" | jq ".main.temp" | cut -d "." -f 1)
-    weather_icon=$(echo "$weather" | jq -r ".weather[].icon")
+    weather_icon=$(echo "$weather" | jq -r ".weather[0].icon")
 
     echo "$(get_icon "$weather_icon")" "$weather_desc", "$weather_temp$SYMBOL"
 fi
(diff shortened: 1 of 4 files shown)


davialexandre:
@x70b1 I couldn't find any explanation on the documentation on why/when that would happen, but it looks like it happens when there's a combination of multiple weather conditions. I only noticed this problem this morning, as it was raining here today but the script was showing it was sunny (I set sunny as the default icon). This is what I was getting from the API:

```json
{
  "coord": {
    "lon": -46.64,
    "lat": -23.55
  },
  "weather": [
    {
      "id": 500,
      "main": "Rain",
      "description": "light rain",
      "icon": "10d"
    },
    {
      "id": 701,
      "main": "Mist",
      "description": "mist",
      "icon": "50d"
    }
  ],
  "base": "stations",
  "main": {
    "temp": 290.43,
    "pressure": 1026,
    "humidity": 82,
    "temp_min": 289.15,
    "temp_max": 292.15
  },
  "visibility": 10000,
  "wind": {
    "speed": 2.6,
    "deg": 150
  },
  "clouds": {
    "all": 90
  },
  "dt": 1528030800,
  "sys": {
    "type": 1,
    "id": 4541,
    "message": 0.0028,
    "country": "BR",
    "sunrise": 1528018931,
    "sunset": 1528057642
  },
  "id": 3448439,
  "name": "Sao Paulo",
  "cod": 200
}

```

x70b1:
Okay I checked that. I got 2 conditions for Sa Paulo too.
This is an interesting case. I would assume there is only one current weather?

But yes, we should use the first one.
Thanks a lot for this work! :sun_with_face:

PR #52, opened 2018-06-04
Fix the script and add some info

unai-ndz:


polybar-scripts/system-nvidia-smi/README.md
```diff
@@ -1,6 +1,6 @@
 # Script: system-nvidia-smi
 
-A script that shows the NVIDIA GPU consumption.
+A script that shows the NVIDIA GPU utilization.
```
x70b1: :+1:

polybar-scripts/system-nvidia-smi/system-nvidia-smi.sh
```diff
@@ -1,3 +1,16 @@
 #!/bin/sh
 
-nvidia-smi --loop=10 --query-gpu=utilization.gpu --format=csv,noheader,nounits | awk '{ print "# $1 %" }'
+# You can use several --query-gpu= values.
```
x70b1: Can we move this info text to the `README.md`?
That would be the right place.

PR #53, opened 2018-06-14
Added info-todotxt

132ikl:


polybar-scripts/info-todotxt/info-todotxt.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+duetoday=$(grep -c "$(date -I)" ~/todo.txt)
+dueweek=0
+
+for i in $(echo {0..7}day); do
```
x70b1: `shellcheck` fails here. for-loops are a bit hard in POSIX. As workaround we could use `while`.

Could you test something like this:
```
weekday=0
while [ "$weekday" -le 7 ]; do
    dueweek=$((dueweek + $(grep -c "$(date -I --date="$weekday")" ~/todo.txt)))
    weekday=$(( weekday + 1 ))
done
```

132ikl:
My solution was hacky in the first place so this works better :smile:

PR #56, opened 2018-06-20
UnseenMail

Danaru87:
My script that display unread email count from multiple accounts.
[UnseenMail Official Repo](https://framagit.org/DanaruDev/UnseenMail)


x70b1:
Why not just use `inbox-imap`?

Danaru87:
The principal reason is "Imap is considered as a less secured protocol by Gmail".

inbox-imap just can be used by only one account at same time.

I'm just requesting a merge because some reddit users want to see my script here ;)

x70b1:
We could add a link [here](https://github.com/x70b1/polybar-scripts#see-also-these-other-user-repositories).
So you can maintain and develop your script in your repo.

Danaru87:
Ok, better solution for me to maintain from my repo.

I cancel the pull request and do another with my repo link in README.

PR #58, opened 2018-07-03
Module to launch the jgmenu from polybar

U00Plinux:


x70b1:
This seems more like a simple configuration. There is no script included.

PR #60, opened 2018-07-05
Added info-stream-countdown

132ikl:
I made this at like 3 AM, so let me know if anything needs to be redone. It's a bit of a mess but there shouldn't be any major issues, barring the 200 MiB memory usage.


x70b1:
Give me some days. I check this as soon as possible.

132ikl:
Thanks. I'm going to disappear for a week on Sunday, so please keep it open until I get back if any changes need to be made.

132ikl:
Fixed a couple silly things, as tends to happen after actually using the module for a day. I should probably start testing these before PRing them but I get too excited lol.

x70b1:
Can you fix the shellsheck issues first?

132ikl:
SC2088 is irrelevant, since the path is just used to fit with the skeleton.
SC2034 is irrelevant, since the echo is commented out by the boilerplate icon to fit with the skeleton.
SC2154 is irrelevant, because showSeconds is used as a easy flip on/off config option at the top, and the issue goes away if you turn on the showSeconds variable.

x70b1:
I merged it. But that scripts spawns a `chrome` every time. Is there no more simple solution for this?

Unfortunately, I don't use twitch. I can't say anything about that.

PR #64, opened 2018-07-21
date-calendar: added script

mjturt:


mjturt:
Yes that's true. I'll make a new pull request.

x70b1:
What is you window manager? I got some problems with `i3wm`. I guess that I should configure floating for `yad`.

132ikl:
@x70b1 `for_window [class="Yad"] floating enable` :)

mjturt:
Bspwm. Yes I totally forget to mention in the README that with tiling window managers Yad needs to be set floating. I'm adding that too.

x70b1:
I work on the files. I will add an info.

mjturt:
Actually yes for pos_y it can be better if it does not depend on mouse position at all.

x70b1:
I merge that. We can fix that later.

PR #68, opened 2018-08-08
Add player-cmus module

tdloi:
This script use `cmus-remote` to control cmus. `cmus-remote -Q` is used to check cmus status (is running, playing, paused)

Do I need to change cmus-remote command  to more friendly name?
Ex: `cmus-remote --next` instead of `cmus-remote -n `

Shellcheck: passed


x70b1:
I will check this in the next days.

PR #70, opened 2018-08-14
System cpu cores temp pure sh

thugcee:
One more script to display CPU cores temperatures. Its features are:
* uses only pure ``/bin/sh`` and ``/bin/cat``
* can change text color when temperatures are above some value.


thugcee:
I'll try again :)

PR #71, opened 2018-08-14
System cpu cores temp pure sh

thugcee:
One more script to display CPU cores temperatures. Its features are:
* uses only pure ``/bin/sh`` and ``/bin/cat``
* can change text color when temperatures are above some value.


x70b1:
Why not the polybar module [temperature](https://github.com/jaagr/polybar/wiki/Module:-temperature)?

thugcee:
hmm.. I've written the script some time ago because internal temperature module wasn't working on two of my computers, but when you mentioned it I've tested it again and now it works great :face_with_head_bandage: 
:ok: let's close this PR :smile:

thugcee:
I've tried to switch from my script to internal temperature module and I've found following problems:
- to display per core temperatures (this is a must have feature on some CPUs which do not have integrated temperature meter) I have to create one configuration section per core, which is cumbersome when configuration is shared on many computers and those computer have different numbers of cores,
- creating separate modules per core means there will be a separator string between them, which doesn't look cool.

So maybe someone will find my script useful as I'm finding it useful in my specific case (shared config, per core meters, different numbers of cores).

The script is dead simple so it should have minimal maintenance burden.

x70b1:
We have already `system-cpu-temppercore`. Does it help?

thugcee:
Yes it helps, but:
- requires `lm-sensors`,
- spawns 6 processes, among those is quite heavy `awk` - it's not negligible on machines like Raspberry Pi,
- doesn't support warning color.

x70b1:
Then you have a very specific use case. That doesn't apply to many people.
We can't build scripts for every case here.

Issue #35, opened 2018-03-16
player-mpris-tail no output


ibrokemypie:
Hey, when trying to use player-mpris-tail I simply get no output in my bar. Playerctl itself correctly shows my players and the current status, but even running the script manually does nothing. I am on arch linux with python up to date.

x70b1:
Can you paste the needed parts of your config?

ibrokemypie:
```
[module/player-mpris-tail]
type = custom/script
exec = ~/.config/polybar/player-mpris-tail.py
tail = true
click-left = ~/.config/polybar/player-ctrl.sh previous
click-right = ~/.config/polybar/player-ctrl.sh next
click-middle = ~/.config/polybar/player-ctrl.sh play-pause
```

x70b1:
And you added `player-mpris-tail` to your bar?
Like `modules-right = player-mpris-tail`

ibrokemypie:
Yes, I did

x70b1:
You are right. I have the same problem here on Fedora. The script run but there is no output.

```
$ python3 -V
Python 3.6.4
```

Unfortunately I do not use a player script. So I have not much experience with it.
Maybe @Cybolic can help. He was the last one who worked on this script. Let's give him some time to answer.

Cybolic:
Maybe it's a case of missing packages; I added a call to `dbus-send`. Could you check if that command is available on your system?

x70b1:
`dbus-send` is available.

I checked that on my Arch device. It works if I install `playerctl` and `playerctl-git`. 
It's a little bit weird. It works on my Fedora now too. But I haven't changed anything.

@ibrokemypie could you find out something?

@Cybolic should we add `python-gobject` as dependency? Otherwise I got an error:

```
python3 polybar-scripts/polybar-scripts/player-mpris-tail/player-mpris-tail.py 
Traceback (most recent call last):
  File "polybar-scripts/polybar-scripts/player-mpris-tail/player-mpris-tail.py", line 8, in <module>
    gi.require_version('Playerctl', '1.0')
AttributeError: module 'gi' has no attribute 'require_version'
```

Cybolic:
Yes, I think that would be a good idea.
Outside of this issue, I'd like to revisit the script at some point and do a rewrite as it's still not great at handling multiple players and can be redone to not actually need `playerctl` (or `dbus-send` for that matter). Unfortunately, I'm swamped with work these days, but I do expect to get on it within a month or so.

[...]

PR #74, opened 2018-09-11
mpris-tail: add configurable output format

avernan:
Add the possibility of passing an argument to the python file to define the output format.

Currently supports tags for icon, artist and title.


avernan:
OK, I've just found out the discussion happening in #35 and the updated script is working for me, as well as taking care of formatting.

PR #75, opened 2018-09-12
Added program-counter plugin.

jtambasco:
Added GUI and process counter plugin that may be of use to others.


x70b1:
Give me some days to check this! :sunglasses:

132ikl:
Why do you have lists of GUI when you could just do a `wmctrl -l | wc -l`? That way users don't have to add programs manually and it will always display as many windows are open.

jtambasco:
I wanted to see the counts of different apps running: i.e. 4 instances of chrome, 2 instances of termite etc.  To achieve this, a glyph is associated with each app, and I check just for that app running.  Also, I didn't really want to know whether some GUI programs are running.

I guess if one wanted to generically monitor the total number of GUIs running, there could just be 1 set glyph for that and then the user wouldn't have to edit anything; however, then the user wouldn't know which individual programs are running.

Also, the plugin monitors processes, and it might be confusing asking the user to specify which process to watch, but not which GUIs to watch.  (Displaying the total process count is probably meaningless?)

If there is interest, I can briefly make another plugin (or modify the existing one) to count the number of GUIs running.

PR #76, opened 2018-09-16
Rewrite, documentation and support for new `truncate-text` arg

Cybolic:
The rewrite that has been tested in #35.


x70b1:
This rocks!

PR #77, opened 2018-09-16
General improvements of the updates-arch scripts

Cybolic:
- Auto-select helper
- Added `yay` helper
- Added fallback for when no known helper is installed


x70b1:
Nice idea. But is it necessary?
I'm not sure. I guess everyone should know which AUR helper is in use.

x70b1:
But adding `yay` is a good idea.

Cybolic:
The problem I was facing with just adding `yay` as a commented out line (which is what I started with), is that the command for getting the amount of changes for both the repo and aur with `yay` is a one-liner. It bypasses a lot of code in the "combined" version, which I was having trouble making into an easily editable version.

x70b1:
We are a bit at the point where we want make it perfect for everyone. But thats not possible.
I want to make the setup as simple as possible. But that doesn't mean that the scripts fit at all points for your own use case. So we should focus on things that are really helpful. 

I guess its the best idea to add a `yay` example.

You need the `pacman-contrib` package. It contains `checkupdates`. Look at the [module page](https://github.com/x70b1/polybar-scripts/tree/master/polybar-scripts/updates-arch).

Is there a way to ask `yay` to show only AUR updates?
Then we could add this as example very easy. 

Also we can add a note about the one for all `yay` command.

Cybolic:
I had put it as `yay -Qu --aur --quiet` to match the other helpers, but both should work.

x70b1:
So `yay -Qau` shows all AUR updates only, right?
And `yay -Qu` shows all updates including AUR updates?

Is there a way to show regular updates only?

x70b1:
I opened an issue: [Jguer/yay/issues/715](https://github.com/Jguer/yay/issues/715)

Cybolic:
As suggested, I created a separate script for `yay` in a new pull request.

x70b1:
Ahh. Not so fast :) 
I will wait for the answer at `yay`. If they introduce a solution, we could very easy fix our scripts.

Cybolic:
Fair enough :) I've submitted another pull request that just adds comment to the current scripts for if/when `yay` fix their output. At that point I'll let you decide which version you prefer :)

Cybolic:
What about the commented use of pacman directly to get the repo updates, is that no good?

x70b1:
`pacman -Qu` and `yay -Qu` are the same thing. They return a list of upgrades. The information for that depends on your local package database. 
Means you have to update (with `pacman -Sy` or `yay -Sy` ) the database before you can see the latest updates.

PR #78, opened 2018-09-21
Arch updates for `yay`

Cybolic:
Independent script for `yay`, including handling its faulty error messages.

As suggested in #77.


x70b1:
We add notes for `yay` in the existing scripts.

PR #79, opened 2018-09-21
Added comments about yay and pacman

Cybolic:
Alternative implementation of `yay` support. Won't actually be viable until `yay` fixes their error output.

See #77 for more info.


x70b1:
@Cybolic: [click](https://www.reddit.com/r/Polybar/comments/9ib9sj/polybarscripts_playermpristail_was_merged/) :wink:

PR #80, opened 2018-09-23
Add standard error redirection

tmwatchanan:
…e in case of no internet connection


x70b1:
It helps to prevent the syntax error messages of `sed`. 
But the script is still broken if there is no connection. 

We should do more work on this script.

tmwatchanan:
Mine works fine. It just does not show rrt number.
https://i.imgur.com/IKPWZXq.png

x70b1:
Yes, thats the point. Something like `Error: I could not ping the target.` Would be better.
Another question is how to handle ping timeouts. 

I will think about it.

Issue #81, opened 2018-09-23
updates-arch-aur,updates-arch-combined: yay displays wrong value


x70b1:
If there is no network connection, yay produces an error output which leads to a wrong value. 
There is an open issue at yay: [#715](https://github.com/Jguer/yay/issues/715)

x70b1:
A workaround for the moment 

```
#!/bin/sh

if ! updates_arch=$(checkupdates 2> /dev/null | wc -l ); then
    updates_arch=0
fi

if ping -n -c 1 -W 1 archlinux.org > /dev/null; then
    if ! updates_aur=$(yay -Qum | wc -l); then
        updates_aur=0
    fi
else
    updates_aur=0
fi

updates=$(("$updates_arch" + "$updates_aur"))

if [ "$updates" -gt 0 ]; then
    echo "  $updates"
else
    echo ""
fi
```

Morganamilo:
Maybe a better work around is just to check the return status. Yay will return non 0 in case of error.

Morganamilo:
Reading the script it seems a little over engineered. Wouldn't something like this work?

```sh
#!/bin/sh
updates="`checkupdates 2> /dev/null; yay -Qum 2> /dev/null`"
if [ -n "$updates" ]; then
	echo "# $(echo "$updates" | wc -l)"
else
	echo
fi
```

Also I don't get why you' want to print an empty line with no updates or prefix the updates with `#  `. Personally I would just echo the number if it exists and that's it.

x70b1:
You're right, checking the return code is a much better idea. 
I fixed that above. Thanks!

Polybar has a "special" behavior. If you want to delete text from it, you need to print nothing. Otherwise, if you just quit the script, the old output would be still on the bar. This means that you could update your system, but the notice on your bar would not be hidden.

The `#` sign is just a placeholder. Most users replace it with symbols from a symbol font.

frebib:
I feel like everyone is over-complicating it, this is what I'm using:

```ini
[module/updates]
type = custom/script
exec = count=$((checkupdates & cower -u) | wc -l); test $count -gt 0 && echo $count || false
interval = 60
format-prefix = "  "
format-prefix-foreground = ${colours.icon}
click-left = i3-msg exec i3-sensible-terminal -r i3blocks-update-term -e yay
```

Of course if you're using yay, swap out `cower -u` with `yay -Qu`

[...]

PR #88, opened 2018-10-05
updates-arch-aur working with trizen, cower and yay

fplucas:


x70b1:
Thanks for your PR!

We had this idea already in #77. I find the way of trying all helpers a bit inappropriate. If everyone can uncomment what they need, it would be a better solution.

At the moment this is not yet possible for `yay` in the same fashion as with the other helpers. But we track that at #81.

PR #94, opened 2018-10-14
Added openvpn-status script

Vincevrp:
This script shows the running openvpn config file.


Vincevrp:
Depends, maybe some people prefer not to show the filename? I would keep `isrunning-openvpn` as a simpler alternative.

x70b1:
Then we could rename your script to `openvpn-filename` and the other one to `openvpn-isrunning`?

Vincevrp:
Maybe `openvpn-config`?

x70b1:
Probably doesn't matter for now. If we notice later that the name is not the best it can be renamed again. This already happened with some scripts.

We can call it `openvpn-status`. Then we can simply add more features when we have them.

Vincevrp:
Alright!

polybar-scripts/openvpn-status/README.md
```diff
@@ -0,0 +1,32 @@
+# Script: openvpn-status
+
+Show the running openvpn config file
+
+![VPN down](screenshots/down.png)
```
x70b1: Please rename it to `1.png` and `2.png`

polybar-scripts/openvpn-status/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+![VPN connected](screenshots/connected.png)
+
+## Dependencies
```
x70b1: I guess we can delete this. 
Everyone should know that he need OpenVPN ifhe want to use this script.

polybar-scripts/openvpn-status/README.md
```diff
@@ -0,0 +1,32 @@
...
+```
+openvpn --daemon --auth-nocache --cd "/etc/openvpn/conf.d" --config "UK-Southampton.ovpn"
+```
+
+## Module
+
+```ini
+[module/openvpn]
+type = custom/script
+label = VPN: %output%
```
x70b1: The `VPN:` text part should be part of the script.

polybar-scripts/openvpn-status/openvpn-status.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
```
x70b1: please use `#!/bin/sh` to be independent from `bash`

Vincevrp:
Fixed it!

PR #95, opened 2018-10-14
pulseaudio-tail: Add default sink

julio-b:
This PR makes it possible to track the default sink.
Right now pulseaudio-tail uses only one sink, which is defined by the $sink variable. But that has some limitations. E.g. users that have more than one sound card may want to display the sink that is active.

### DIFF
--- a/polybar-scripts/pulseaudio-tail/pulseaudio-tail.sh
+++ b/polybar-scripts/pulseaudio-tail/pulseaudio-tail.sh
@@ -1,6 +1,11 @@
 #!/bin/sh
 
-sink=0
+### Leave this variable unset if you want to use the default sink
+#sink=0
+
+get_default_sink() {
+    sink=$(pacmd list-sinks | sed -n '/\* index:/ s/.*: //p')
+}
 
 volume_up() {
     pactl set-sink-volume $sink +1%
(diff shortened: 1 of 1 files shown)


132ikl:
Fix the [shellcheck issues](https://travis-ci.org/x70b1/polybar-scripts/builds/441370249?utm_source=github_status&utm_medium=notification) first so your PR is able to be accepted (at x70b1's discretion).

julio-b:
Sorry about that, it should be ok now.

x70b1:
My other question: Why not the [`pulseaudio`](https://github.com/jaagr/polybar/wiki/Module:-pulseaudio) of polybar? It supports a default sink already. This script was written at a time when no pulseaudio support was availiable.

julio-b:
Correct. To be honest the only reason i am currently using this script is because pulseaudio module can't detect if headphones are plugged in.
But as you said, the module is by far better than this bash script and everyone should use the module instead. I added headphone detection feature here jaagr/polybar#1478. If it gets merged i think you should remove pulseaudio-tail and pulseaudio-simple because they are useless and very slow.
So feel free to reject this PR

x70b1:
I have no plans to reject working scripts. 
Maybe there will come a day when the polybar modules won't work anymore. The scripts are then a welcome workaround.

132ikl:
Maybe add a disclaimer in the pulseaudio-simple and pulseaudio-tail scripts to redirect users to the real pulseaudio module? I didn't know this was a thing either, that information would've been useful.

x70b1:
I added a note about the existing `pulseadio` module.

x70b1:
A lot of time has passed. But I worked a bit on it. There was a race condition on the version of @julio-b.
Can somebody confirm that this works?

Sad-Soul-Eater:
> Can somebody confirm that this works?

All works fine for me.

PR #97, opened 2018-10-18
Make player-moc use filename if no song name

chrjen:
Not every file necessarily has the song field set to anything. In those
cases to avoid empty output, e.i. " - ", the filename is printed instead.


chrjen:
Yes seem to be working fine for all three cases.

x70b1:
Thanks!

PR #100, opened 2018-11-14
removed unnecessary lines

py-radicz:
unnecessary imports removed


132ikl:
Range should be kept for users who want to configure the script to view more than one entry.

py-radicz:
now you can get what feed you want, by modifying url and number of entries variables (less painfully for less experienced users)

132ikl:
nice!

PR #101, opened 2018-11-17
Add anyconnect-status

kremlinkev:
* Adds a simple script for cisco anyconnect client status


polybar-scripts/anyconnect-status/vpnstatus.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
```
x70b1: Could you use `/bin/sh` and fix all `shellsheck` errors? I don't see why bash is explicitly needed here.

polybar-scripts/anyconnect-status/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+Like that glyph? It's 賓 from [FantasqueSansMono](https://github.com/ryanoasis/nerd-fonts/tree/master/patched-fonts/FantasqueSansMono/Regular/complete)
+
+## Module
+
+```ini
+[module/vpn]
+type = custom/script
+exec = ~/bin/vpnstatus.sh
+format-prefix = "#1 "
```
x70b1: We should move this icon inside the script. This allows us to change the icon in different situations. For example colors like red and green depending on the status.

polybar-scripts/anyconnect-status/README.md
```diff
@@ -0,0 +1,37 @@
...
+shots? Here's the exact script:
+
+```bash
+#!/usr/bin/env bash
+
+IFACE=$(ifconfig | grep cscotun | awk '{print $1}')
+
+if [[ $IFACE == "cscotun"* ]];
+then
+    echo "%{u#55aa55}$(ifconfig cscotun0 | grep inet | awk '{print $2}' | cut -f2 -d':')%{u-}"
```
x70b1: Add this line as a comment in your script, so that everyone can simply uncomment this part if he wants colors. Otherwise we have a second version of the same script in our `README.md`. Maybe not the best idea.

polybar-scripts/anyconnect-status/README.md
```diff
@@ -0,0 +1,37 @@
...
+else
+    echo '%{F#FF0000}%{u#FF0000}VPN DISCONNECTED%{u-}%{F-}'
```
x70b1: Look at line 21.

kremlinkev:
@x70b1 Comments addressed, thanks for the feedback.

PR #105, opened 2018-11-27
player-mpris-tail: Only print status from current player

Cybolic:
Currently the script prints the status of each player as soon as it has it. This pull request changes that so a status is only printed once all players have been read and the most likely current one detected (or none have been).
This fixes the issue where the status is always stopped when running certain players in the background.


x70b1:
I broke it ..

x70b1:
Awesome!

PR #106, opened 2018-11-27
player-mpris-tail: Print update when the current player is no longer current

Cybolic:
Currently, when the current player is stopped/paused, the status of the next player in line is not printed; this fixes that.


x70b1:
Oh, I forgot this ..

PR #114, opened 2019-01-07
info-tmux-sessions: add a tmux session tracking module to polybar

guygastineau:
Hey, I just started using polybar, and I really appreciate this repo.

I couldn't find the tmux module I wanted, so I made it and I thought I would share it here.

Let me know if everything is in order.


polybar-scripts/info-tmux-sessions/README.md
```diff
@@ -0,0 +1,21 @@
+# Script: info-tmux-sessions
+
+Reports the state of tmux sessions on the local machine.
+
+* with active sessions
+![info-tmux-sessions](screenshots/1.png)
```
x70b1: The font is a bit small, maybe we need better screenshots

polybar-scripts/info-tmux-sessions/README.md
```diff
@@ -0,0 +1,21 @@
...
+* with no server running
+![info-tmux-sessions](screenshots/2.png)
+
+## Dependencies
+
+* `tmux`
```
x70b1: we could delete this line, I guess its obviously that `tmux` is required

polybar-scripts/info-tmux-sessions/info-tmux-sessions.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
```
x70b1: Is there a reason why we should not write a shell independent script with `#!/bin/sh`?
The most of the scripts in this repo are that. We could work on this.

polybar-scripts/info-tmux-sessions/README.md
guygastineau: I thought that might be a concern.

I'll set up a secondary bar and get clearer screenshots tomorrow.

polybar-scripts/info-tmux-sessions/README.md
guygastineau: I figured as much.  Should I remove the dependencies header too?

polybar-scripts/info-tmux-sessions/info-tmux-sessions.sh
guygastineau: I suppose I did this out of habbit.

Also, I'm not sure that posix would like how I check for the exit value 1 that comes back when there are no sessions.

I am willing to change it to /bin/sh.

I guess it makes my disclaimer that I've only tested it on the latest version of bash all the more important.

guygastineau:
Thank you for those reviews.

I'll get it done tomorrow ;)

polybar-scripts/info-tmux-sessions/README.md
x70b1: Yes ..

polybar-scripts/info-tmux-sessions/info-tmux-sessions.sh
x70b1: I would love to see this and other users too I guess :sunglasses: 
Write a comment if you need help.

guygastineau: Oh, you mean have it tested on sh ?

My Arch system has a symlink to /bin/bash located at /bin/sh.  I might be able to test this on a server that isn't Arch though.

guygastineau: I have now tested it with the `--posix` flag set, and all seems well.

I do not know if this is rigorous enough testing.  Let me know if you have extra ideas for verifying it's compatibility with other shells.

I'll commit an update to the shebang when I've made the new screenshots.

polybar-scripts/info-tmux-sessions/README.md
guygastineau: removed

polybar-scripts/info-tmux-sessions/README.md
guygastineau: Much bigger screenshots now.

polybar-scripts/info-tmux-sessions/info-tmux-sessions.sh
guygastineau: I have confirmed this does not work with dash.

guygastineau: dash doesn't have `[[`

guygastineau: After some more research I believe this will work with the more advanced members of the korn shell family, but I do not expect this to work with the plain Bourne shell.

guygastineau: It is possible to rework it for compatibility with dash, etc.. , but it would involve external commands like `sed`.

Overall it would be harder to read, and probably slower due to the workarounds and external programs.  Right now it takes less than 1/100 of a second on my machine.  Should I add this as a disclaimer?  People on debian will get dash if it is invoked with `/bin/sh`, and then it will be broken for them.

guygastineau: Also a problem for void users.

x70b1: You can use `shellcheck` to see what happens. 
The main problem is that you use arrays. POSIX can't handle arrays. The qestion would be: Is it really needed?

guygastineau:
I should be able to replace the arrays at the expense of readability.  I have to make sure everything will work with `[` instead of `[[` too I suppose.

guygastineau:
Hey, sorry to leave this hanging for so long.

I suppose I am loath to say, that with school starting again I don't have the time or interest in making this script work with sh.

The solution would be much uglier that the current one, and executing this script in bash does not harm system resources.

If it is left as bash, then the shebang should be followed with `/usr/bin/env bash` for portability across systems with bash.

If you are not interested in including this script as a bash only module, then I suppose this PR should be closed.

Thank you for your time.

x70b1:
No problem. 
If you leave it open I rewrite it in the next days. :+1:

guygastineau:
Thanks, I think this is a handy tool, and I am very thankful that you are willing to tackle the `sh` compatibility.

If I come up with any new tools I'll try to make them `sh` compatible before submitting a PR. 😉

x70b1:
Thanks for that!

PR #119, opened 2019-01-22
Encrypted password with gpg keys.

AKXX:


132ikl:
Maybe this should be an option instead of a mandatory change, by either putting it in a separate script or by adding options within the script? That's up to x70b1 though

x70b1:
Thx @AKXX!

PR #123, opened 2019-02-11
Add info-hlwm-workspaces

patrick96:
I wasn't sure about the colors since they're kind of essential but we should still let the user customize them. I opted for not using any colors and instructing the user to add them themselves.


patrick96:
Finally, my first PR here ;)

x70b1:
It was time for that.

PR #130, opened 2019-02-22
schedule display module

DJSiddharthVader:
module to display current schedule entry from csv file


x70b1:
Due to the special and rare purpose, I would suggest you should outsource and maintain the script in a separate repo.

I would link it in the README.md if you are still interested.

PR #131, opened 2019-02-22
yad popup calendar - add auto bottom detect & out of screen position fix

Sad-Soul-Eater:
before after calendar

### DIFF
--- a/polybar-scripts/popup-calendar/popup-calendar.sh
+++ b/polybar-scripts/popup-calendar/popup-calendar.sh
@@ -3,7 +3,7 @@
 BAR_HEIGHT=22
 YAD_WIDTH=200
 YAD_HEIGHT=200
-DATE=" $(date +"%e %B")"
+DATE="$(date +"%a %d %H:%M")"
 
 case "$1" in
 --popup)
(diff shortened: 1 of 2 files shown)


polybar-scripts/popup-calendar/popup-calendar.sh
```diff
@@ -1,27 +1,36 @@
-#!/bin/sh
+#!/bin/bash
```
x70b1: Please use `#!/bin/sh` and fix all shellcheck errors.

Sad-Soul-Eater: With `#!/bin/sh` shellcheck reports warnings [SC2039](https://github.com/koalaman/shellcheck/wiki/SC2039) in lines 14, 16, 23. But it works properly on my system. Should i try to fix it or `#!/bin/sh` will be enough?

x70b1: Yes, please fix the errors. 
I guess it fails at your conditions. Try to use `if [ .. ]; then` instead of `if (( .. )); then`.

x70b1:
Good work @Sad-Soul-Eater!

PR #136, opened 2019-02-27
Add module: vpn-wireguard-wg

Sad-Soul-Eater:


polybar-scripts/vpn-wireguard-wg/vpn-wireguard-wg.sh
```diff
@@ -0,0 +1,66 @@
...
+
+is_exists() {
+    if [ ! -f $CONFIG_PATH ]; then
+        echo "$DISCONNECTED_ICON Config file not found"
+        exit 0
+    fi
+}
+
+check() {
+    CONFIG_NAME=$(basename "${CONFIG_PATH%.*}")
```
x70b1: Your last commit was a nice idea. 
You could move this two lines outside this function and aftr that you can simplify the connected condition too.

Sad-Soul-Eater: I don't think it's a good idea because connected condition is changing in `toggle()`, so if we don't want to get double delay for status update, we need recheck connection and only after that call `status()`.

x70b1: Ah, yes. You are right.

Sad-Soul-Eater: > Ah, yes. You are right.

I thought, that polybar is updating module text on click action... It doesn't - https://github.com/jaagr/polybar/issues/786
But as I understand this feature will be added in the next update - [click me](https://github.com/jaagr/polybar/projects/1#card-5907583)
So maybe left that as is instead of updating on next polybar release?

polybar-scripts/vpn-wireguard-wg/vpn-wireguard-wg.sh
```diff
@@ -0,0 +1,65 @@
...
+    exit 0
+fi
+
+CONFIG_NAME=$(basename "${CONFIG_PATH%.*}")
+
+check() {
+    WG_RESULT=$(sudo wg show "$CONFIG_NAME" 2>/dev/null | head -n 1 | awk '{print $NF }')
+
+    if [ "$WG_RESULT" = "$CONFIG_NAME" ]; then
+        CONNECTED=true
```
x70b1: Lets work with a return value instead of writing a var.

polybar-scripts/vpn-wireguard-wg/vpn-wireguard-wg.sh
```diff
@@ -0,0 +1,65 @@
...
+
+    FULL_CONFIG_PATH="$(readlink -f "$CONFIG_PATH")"
+
+    if $CONNECTED; then 
+        sudo wg-quick down "$FULL_CONFIG_PATH" 2>/dev/null
+    else
+        sudo wg-quick up "$FULL_CONFIG_PATH" 2>/dev/null
+    fi
+
+    status
```
x70b1: This output will never be seen.
We can delete the command?

Sad-Soul-Eater: There is a way to rewrite script totally without functions, but when output on click event been realized in polybar, will back to this. What's better?

x70b1: Functions are not bad if the used more than one time.

x70b1:
Sorry, a bit late.
I changed a lot. Can you check it?

x70b1:
Looks good, thanks!
Should we leave it in this condition?

Sad-Soul-Eater:
> Looks good, thanks!
> Should we leave it in this condition?

Yep 👌

PR #137, opened 2019-03-03
Fixes for fedora update checker

kvnbai:
FIXES:
* dnf upgrade needs sudo to be executed
* add space to remove the matching 'dependencies' string in $integer


x70b1:
Thanks!

PR #141, opened 2019-03-07
Tickers for BTC\USD and ETH\USD pairs

Corenit:
both based on https://github.com/x70b1/polybar-scripts/tree/master/polybar-scripts/ticker-btceur


x70b1:
Thanks for your work on it. But in general it is not a significant change compared to the existing scripts. That's why I won't pull it. I want to prevent duplicate work. But everyone is able to customize the scripts.

PR #153, opened 2019-04-05
Added polybar node-version

zemmsoares:
Python script to compare latest available Node.js version to the installed


x70b1:
Why isn't that written in nodejs? :rofl:

zemmsoares:
Hehe it would be easier for me scraping with node :rofl: ,
but i'm trying to learn python, and i think this is a good way to start :+1:

PR #156, opened 2019-05-05
DualShock 4 Battery level

d3rrial:
This script shows the current battery level of the DualShock 4 Controller in 10% increments.
I'm not good with Bash, so someone please check if I forgot to escape something important or did something silly.


x70b1:
Good idea!
Can you post an output of `/sys/class/power_supply/sony_controller_battery_*/capacity` ?
So that i can improve the script a bit.

d3rrial:
The output is literally just an integer between 0 and 100
I'm not sure if I can describe it in any more meaningful way, sorry

x70b1:
No thats enough! :)

x70b1:
@d3rrial Thanks for the scripts!

PR #157, opened 2019-05-07
Spotify panel for polybar

d3rrial:
I'm sure there are more efficient ways to do this, I've basically used the first thing that worked for me, with no attention to elegance.

I apologize for using font awesome as well as custom colors. The symbols used are all listed.
As for the colors, they're set but easily accessible/changeable.


d3rrial:
Actually player-mpris-tail does this much better, feel free to discard this, I'm not sure I'm adding anything in particular.

x70b1:
@d3rrial Thanks a lot for your work!
You're right `player-mpris-tail` is already a better solution for this.

PR #161, opened 2019-06-08
Make player-mpris-tail able to handle position in the current track

Ekdohibs:
The `{position}` tag can now be used to get the progress on the current track and display it. For now, only display as a time is supported, but I can add display as a progress bar if it people wish so.
The code tries not to make too frequent calls to dbus, and is disabled when not using it in order to save battery.

### DIFF
--- a/polybar-scripts/player-mpris-tail/player-mpris-tail.py
+++ b/polybar-scripts/player-mpris-tail/player-mpris-tail.py
@@ -130,6 +130,12 @@ class Player:
             'title'  : '',
             'track'  : 0
         }
+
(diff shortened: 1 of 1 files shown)


x70b1:
@Cybolic Any thoughts?

Cybolic:
I haven't tested it, but it looks good :)

Ekdohibs:
I commited a fix to handle the mpris2 `Seeked` signal that is sent when the position in the current track changes, as well as a fix with players that improperly send Rate-change events instead of play/pause (this seems to happen with vlc).

PR #162, opened 2019-06-22
Added pavolume module

marioortizmanero:
I've just added a very handy pavolume module. It looks simple but it's super useful, check the README for more.


x70b1:
What is the real benefit over [pulseaudio-tail](https://github.com/polybar/polybar-scripts/tree/master/polybar-scripts/pulseaudio-tail) or polybars [pulseaudio module](https://github.com/polybar/polybar/wiki/Module:-pulseaudio)?

marioortizmanero:
Well this script can switch between sinks very easily (and with a blacklist), which is why started this in the beginning. Without it you'd have to open pavucontrol, go to output devices, change the default one, go to audio sources, move all of them to the new device... Or do it with pacmd which is even worse.

It also has support for osd on kde, a volume cap and a couple other things

juansquintero:
It works perfectly

marioortizmanero:
So, have you considered adding it? Maybe I should put it in a repo of my own and add a link in the README instead?

x70b1:
I would prefer to extend the existing script instead of adding a new one.
A separate repository would make sense. But thanks a lot for your contribution to polybar-scripts.

marioortizmanero:
No problem! I've added the link to the repo in case you want to add it but you can just close the PR otherwise.

PR #165, opened 2019-07-01
Simplifies positioning of window in calendar popup script

Grannath:
This a) simplifies things by a lot, and b) works on multiple monitors.

Before, positioning would fail completely for me. I have a small screen to the left, and my main screen to the right. The window would open way to the left, and too far down. The mouse parameter works just fine.


Sad-Soul-Eater:
Not so fine

2019 07 30-17 35 17 screenshot
2019 07 30-17 34 57 screenshot

x70b1:
This is more a dirty fix than a good solution. 
I opened an issue #171 for an inprovement.

PR #166, opened 2019-07-11
updates-void-notify: add script

tuxliban:
Hello, I have made a script for the distribution Void Linux which in addition to showing the available updates in the pobybar bar also shows what updates are available through a pop-up message.


x70b1:
Thanks a lot for your pull request. 

I think notifications are a bit out of scope here. Because they don't have much to do with the polybar anymore. That's why I wouldn't continue with it.

PR #167, opened 2019-07-22
updates-fedora.sh: fix for working on different locales

fcomida:


x70b1:
Looks good, what is the output if you dont set the `ENV`?

fcomida:
@x70b1 no output at all. My locale is it_IT and **dnf** outputs in italian: "Upgrade" --> "Aggiornamenti"

x70b1:
Yes, I understand. Thats the problem if we grep something.
Thanks for that fix! :+1:

PR #177, opened 2019-09-09
Check Wifi-ESSID for WIFIonICE

luerhard:
This ensures that the output is removed if the network is no longer
reachable.
The script is enclosed with an additional if-clause that does a iwgetid -r request, which outputs the current ESSID from that wifi network one is connected to.

### DIFF
--- a/polybar-scripts/info-wifionice/info-wifionice.sh
+++ b/polybar-scripts/info-wifionice/info-wifionice.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-if [ "$(iwgetid -r)" == "WIFIonICE" ]; then
+if [ "$(iwgetid -r)" = "WIFIonICE" ]; then
     wifionice=$(curl -sf https://iceportal.de/api1/rs/status)
 
     if [ "$(echo "$wifionice" | jq .connection)" = "true" ]; then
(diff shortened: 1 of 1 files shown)


x70b1:
A simple 
```
else
    echo ""
fi
```
should already fix all this?

luerhard:
The important part is the one at the beginning.
```
if [ "$(iwgetid -r)" == "WIFIonICE" ]; then
#all the rest of the script
else
    echo ""
fi
```

At least for me the supresses the `parse error: Invalid numeric literal at line 1, column 7` which occurs in the current master-version of this script when I am not connected to a WIFIonICE network. And it ensures that the messages won't stay in my polybar forever when I disconnect from such a network.

x70b1:
For me, the use case is to remove the script when I'm not using a train. But that's another thing.

At least we should fix the shellcheck error.

luerhard:
Okay, I do not want to mess with my polybar configuration every time I step on a train :) This is what I implemented and I wanted to share that with you, because I really like the script. Maybe you can do something useful with it. As you like.

x70b1:
If you fix the shellcheck I would pull it :+1:
(replace  `==`)

luerhard:
You mean like that? Sorry, I am not really used to shell scripting.

x70b1:
Awesome :sunglasses:

PR #178, opened 2019-10-01
Added Quiet OpenVPN-Status Module

Skoetting:
In the quiet Module you will not see anything if there is no active VPN-Connection


GaffaOverflow:
done

PR #179, opened 2019-10-01
Added Info public ip

Skoetting:
Added the Module info-public-ip


x70b1:
Nice idea! Thanks :sunglasses:

PR #180, opened 2019-10-02
fixed weather variable scope

s-kaczmarek:
Hi, script: openweathermap-detailed.sh was invalid - it was trying to echo some values taken from "weather" variable (containing JSON from OpenWeather API), but "weather" variable was not accessible in it's scope.

I have fixed it by declaring empty weather variable at the beginning of script.

### DIFF
--- a/polybar-scripts/openweathermap-detailed/openweathermap-detailed.sh
+++ b/polybar-scripts/openweathermap-detailed/openweathermap-detailed.sh
@@ -28,6 +28,7 @@ KEY=""
 CITY=""
 UNITS="metric"
 SYMBOL="°"
+weather=""
 
 API="https://api.openweathermap.org/data/2.5"
 


x70b1:
On which line did you try to access the variable?

s-kaczmarek:
Last one, script was not printing what it was supposed to.

x70b1:
I am not sure if I understand your problem.
If there is no weather information, for whatever reason, then nothing is printed. `weather=""` doesn't change anything.

s-kaczmarek:
Yesterday, I decided to add this module to my polybar config. It was not working, so I decided to execute this script from terminal. It didn't printed anything, so I decided to look into code. I'm not bash expert, but what I noticed, was that there is assignment of data from api to "weather" variable in first if else block. 

There is also second if block, that is validating whether string assigned to this variable is not empty. I came with idea, that maybe it can not validate it, because this variable is not visible for this second block. I made declaration of this variable at the beginning of script and now it works perfectly. Variable is declared at the beginning, gets data from api in first block and can be validated in second block, because it reaches it's scope.

Maybe, you could try it yourself- execute this script, see what is returned to console and then apply my fix.

x70b1:
The point is it doesn't show something if there was a mistake somewhere before. 
For example, if you didn't specify an API key. Then the output is empty. Maybe that was the error?

I tested the script from the repo. It does what it should.

s-kaczmarek:
Yes, you are right. I think, I have miss pasted api key at first, thought
it is ok, so I was trying to fix it other way. Probably, I pasted api key
once again and added this additional variable at the same time and I
thought, that it was the fix.

I tested it myself once again today, commented out my additional variable
and it is still working, so it was problem with api key for sure.

pt., 4 paź 2019 o 09:14 x70b1 <notifications@github.com> napisał(a):

PR #183, opened 2019-10-09
player-mpv: shows current music track or movie title with underline as progress bar

vyachkonovalov:


polybar-scripts/player-mpv/mpv-control.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
```
x70b1: Can we include the function of this script into the python script as command line parameters? Then we only have one script file.

vyachkonovalov: Technically we can, but that leads to code complexity (btw, player-mpv.py code already too branchy). Also, the reason mpv-control.sh is a separate bash script is to load it as quick as posible. For operations like time-pos (seek forward/backword), which by default happends on mouse scroll, the speed is important.

x70b1: `player-mpris-tail` is 4 times longer, I don't think it's too complex.

Is the speed really noticable?

vyachkonovalov: Will check soon.

x70b1: Would be perfect! :+1:

vyachkonovalov: The speed isn't noticable. I didn't know python is fast😃

x70b1: Speed is relative. We work with scripts here. They are fast and easy to customize. If we would need execution speed, we would do things in C.  :+1:

polybar-scripts/player-mpv/readme.md
```diff
@@ -0,0 +1,40 @@
+# Script: player-mpv
```
x70b1: Lets rename the script and the directory to `player-mpv-tail`.

polybar-scripts/player-mpv/readme.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Configuration
+
+### 1. *~/.config/mpv/mpv.conf*
+```
+input-ipc-server=/tmp/mpvsocket
+```
+or run mpv with `--input-ipc-server=/tmp/mpvsocket` option.
+
+### 2. Use narrow font like 'Fira Sans Compressed'
```
x70b1: Remove the font suggestion.

x70b1:
Thanks @vyachkonovalov!

Issue #188, opened 2019-10-16
player-mpris-tail compatibility with Tauon Music Box


Taiko2k:
Using player-mpris-tail with my player Tauon Music Box does not seem to function when player-mpris-tail has already started, but does when reloading. Ref issue https://github.com/Taiko2k/TauonMusicBox/issues/187

This ***does*** seem to function correctly with other MPRIS servers, though I've not had this issue with other MPRIS clients.

My server implementation is here. https://github.com/Taiko2k/TauonMusicBox/blob/master/t_modules/t_dbus.py. Could someone please take a look and see if I'm doing something wrong here? Or is this an issue with player-mpris-tail?

x70b1:
Well, we have some issues open. They all relate to the script `player-mpris-tail`. So it's quite possible that the problem is the script.

The script was contributed by the community, it is from @Cybolic. So I can't say much about it, because I never looked deeper into it. We can only leave this issue open for now, maybe someday there will be someone who wants to solve these problems.

Cybolic:
I can confirm that this is a bug/missing feature in player-mpris-tail.py.
The dbus list is only gone through on init and only clients that generate a `NameOwnerChanged` signal are picked up after that.
I'm not sure exactly which event to listen to in order to detect a new client that doesn't trigger `NameOwnerChanged`, but I'll take a look at it within the next 24 hours.

PR #189, opened 2019-10-18
Get player updates without `NameOwnerChanged` #fixes 188

Cybolic:
Some players (Tauon so far) don't seem to trigger `NameOwnerChanged` when they start, so the signal handler for `PropertiesChanged` isn't attached initially.
This PR listens to all signals on `/org/mpris/MediaPlayer2` and specifically catches any that wouldn't be caught due to the missing signal handler and forwards them.
This only seems to happen on the first player change, so no further changes should be necessary.


Cybolic:
This should fix #188

PR #191, opened 2019-10-18
Feature/player mpris tail fix for closed players

### LINKED ISSUE #168: mpris-player-tail does not update after closing the player
**OS**
Arch Linux
**Steps to reproduce:**
1.  Play anything (with SMPlayer in my case)
2. Close the player
3. The module remains playing state and will not disappear

Cybolic:
This should fix #168

### DIFF
--- a/polybar-scripts/player-mpris-tail/player-mpris-tail.py
+++ b/polybar-scripts/player-mpris-tail/player-mpris-tail.py
@@ -70,8 +70,14 @@ class PlayerManager:
         if owner in self.players:
             self.players[owner].disconnect()
             del self.players[owner]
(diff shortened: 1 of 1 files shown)


x70b1:
Wow, good work!
I will pull it later. Thanks a lot! :sunglasses:

x70b1:
@Cybolic everything is merged now. Thanks for your work on this topics. It seems your script is very popular. At the same time there are many different players with different behavior.

I am happy to see that not all scripts here are only from me but also from others.

Cybolic:
@x70b1 Thanks, I just wish I had time more often to fix issues :)

PR #193, opened 2019-10-21
Correctly parse multiple artist metadata

RealOrangeOne:
Looks like there's a typo in https://github.com/polybar/polybar-scripts/pull/192, which results in an exception being thrown consistently. 

I don't know what situation it happens, but it's reliable, and this fixes it.


Cybolic:
Oops! You're very right @RealOrangeOne; thanks for the fix!

x70b1:
Merged. Thx guys!

PR #195, opened 2019-10-31
Add rua support to updates-arch-[aur|combined]

AblingerOscar:
Adding support for rua as [it now supports only listing packages](https://github.com/vn971/rua/issues/86).


x70b1:
Looks good, thanks!

PR #199, opened 2019-11-09
Average Frequency over Cores

d3rrial:
I don't know how useful this is in practice, but I found that just showing the frequency of a single core is often pointless (as it might be one your scheduler ignores if not much is going on), so I decided I'd want to see the average frequency.

There is one other script here that shows cpu frequency, but that is for core 0, which in my case is basically pegged to 2ghz by my governor because my scheduler doesn't assign anything to it.

I hope I followed the styleguide better this time. Sorry if I missed anything :)

This alternative I frankensteined displays the frequency of your highest clocked core:
```bash
#!/bin/bash
m=0;
while read -r i
do
    if (( $(echo "$i > $m" |bc -l) ))
    then
        m=$i
    fi
done < <( awk '/MHz/ {print $4}' < /proc/cpuinfo )
echo "scale=2; $m / 1000" | bc | awk '{print $1" GHz"}'

```


d3rrial:
To be honest I don't know how to fix what travis wants me to do right now.

x70b1:
No worries. The count of commits isn' a problem here. I will merge it to a single one.

I will look into it, but I need some days. 
Thanks for your PR! :+1:

polybar-scripts/system-cpu-avg-freq/system-cpu-avg-freq.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
x70b1: use `/bin/sh` here and fix the shellcheck errors

PR #200, opened 2019-11-28
[player-mpris-tail] Fix dbus.connection exceptions

### LINKED ISSUE #194: player-mpris-tail: intermittent dbus connection problem
Just trying to set up MPRIS info, using "Output using formatters and Polybar action handlers".

Sometimes when I open `mpv` (using mpv-mpris (mpris.so)) I get `ERROR:dbus.connection:Exception in handler for D-Bus signal:` appearing in the bar. Changing the album track in `mpv` can trigger this and so can clicking on an action handler.

It seems to happen at random sometimes, though watching closely and there is always a dbus error flashing up for a fraction of a second before the song title text etc appears when it does work.

dshoreman:
When initially running player-mpris-tail.py, it displays the current
track being played without issue. However when the track is changed
(whether that's skipping to the next, or restarting current) [two
exceptions] get thrown causing the script's output to be replaced with
`ERROR:dbus.connection:Exception in handler for D-Bus signal:`.

The first—`Playback position not available`—is fixed by catching
`dbus.exceptions.DBusException` so we can set a sane default for
`speed_us` when the track's playback position isn't available.

The second—introduced with the fix for #188—is caused by listening for
all signals, some of which don't send the `properties` and `signature`
arguments which are required by the `onMPRISSignal` handler.

As we're only taking action on the `PropertiesChanged` signal, we can be
explicit about that when registering the handler such that only signals
with the arguments we need will ever try to call our handler function.

To reflect its more specific purpose, I've also renamed the handler to
`onChangedProperties`, keeping it in line with `onOwnerChangedName`
while also avoiding any confusion with `player.onPropertiesChanged()`.

---

Tested in mpv, Rhythmbox, and also Tauon Music Box to make sure #188
didn't get broken by the changes.

Fixes #194.

[two exceptions]: https://github.com/polybar/polybar-scripts/issues/194#issuecomment-559527744

### DIFF
--- a/polybar-scripts/player-mpris-tail/player-mpris-tail.py
+++ b/polybar-scripts/player-mpris-tail/player-mpris-tail.py
@@ -41,23 +41,23 @@ class PlayerManager:
 
     def connect(self):
         self._session_bus.add_signal_receiver(self.onOwnerChangedName, 'NameOwnerChanged')
-        self._session_bus.add_signal_receiver(self.onMPRISSignal, path = '/org/mpris/MediaPlayer2',
-            sender_keyword='sender', member_keyword='member')
+        self._session_bus.add_signal_receiver(self.onChangedProperties, 'PropertiesChanged',
+                                              path = '/org/mpris/MediaPlayer2',
+                                              sender_keyword='sender')
 
-    def onMPRISSignal(self, interface, properties, signature, member = None, sender = None):
-        if (member == 'PropertiesChanged'):
-            if (sender in self.players):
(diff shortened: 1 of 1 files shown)


x70b1:
Thx for this @dshoreman.
Let's see what @Cybolic says.

Cybolic:
Lovely changeset and great explanation of what you've changed! I'm all for merging this :)

x70b1:
Thx for this @dshoreman!

PR #202, opened 2019-12-03
Add updates-flatpak

j2ghz:
I couldn't find a better way.
If there are no updates, `flatpak update` output has 2 lines. 
If there are some, it's 5 + number of updates.

I also made build.sh executable


x70b1:
> I also made build.sh executable

Whats the reason for this? 
Travis works without it.

The rest looks nice! I have to test it first as I don't use flatpak. Give me some time for it.

j2ghz:
> > I also made build.sh executable
>
> Whats the reason for this?

I cloned the repository locally, made my changes, and when I wanted to run `build.sh`, I had to run `chmod +x build.sh` before. It's easier for contributors if it's executable right from the start.

x70b1:
Just use `sh build.sh`?

j2ghz:
The most common way to execute a script is to do `./build.sh`, and that doesn't work if the file is not executable.

x70b1:
I don't want to discuss here what the common way is for the execution.
At least it should be in its own pull request.

polybar-scripts/updates-flatpak/README.md
```diff
@@ -0,0 +1,18 @@
+# Script: updates-flatpak
+
+A script that shows if there are updates for FlatPak.
+
+
+## Dependencies
+
+* `flatpak` to run `flatpak update`
+
```
x70b1: I guess we can delete this section. This is very obvious.

polybar-scripts/updates-flatpak/updates-flatpak.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+if ! updates=$(echo "n" | flatpak update 2> /dev/null | wc -l ); then
```
x70b1: We should think about using `flatpak update --noninteractive`? This might be cleaner.

j2ghz: I tried using that and the problem is it just runs the update. `flatpak update --help` doesn't list any way to only print updates.

x70b1: Mh k, that's annoying. I had tested it, but I have not tried it with updates.

x70b1:
@j2ghz Thanks for this! :+1:

PR #203, opened 2019-12-08
Added threshold and font awsome icons.

Liupold:
why update when nothing is happening?. Added a threshold for that (separate for upload and download). As there is nothing happening at ideal interval can be decreased from 10 to 1. (makes it much better and snappier).


Liupold:
ya, the battery issue was prominent.

PR #205, opened 2019-12-25
Specifying a date string in options

Xallt:
The popup-calendar script has a default date string, but specifying a different one by editing the script isn't very convenient, so it's better to set it with an argument


x70b1:
It just makes the script more complex. Also, this value would have to be documented. If the script in the repo is not enough for your purpose, you are free to customize it for yourself. 

But thanks anyway for your work and your PR. :+1:

PR #214, opened 2020-02-24
Tidied & Optimized Battery Code

terminalforlife:
Were quite a lot of redundancies, all of which I believe I've addressed. Hope these help. I'm happy to work on some of the other code, as I can see a lot which could be tweaked/optimized.

This is a really cool project. I'm not a Polybar user myself, but I'm more than familiar with shell.

I see there's a few other PRs gone unnoticed for a while. I hope this project is still very much alive!

### DIFF
--- a/polybar-scripts/battery-combined-shell/battery-combined-shell.sh
+++ b/polybar-scripts/battery-combined-shell/battery-combined-shell.sh
@@ -28,7 +28,7 @@ if [ $AC -eq 1 ]; then
     if [ $BatPerc -gt 97 ]; then
         printf '%s\n' "$Icon"
     else
-        printf '%s %d%%\n' "$Icon" "$BatPerc"
+        printf '%s %d %%\n' "$Icon" "$BatPerc"
     fi
 else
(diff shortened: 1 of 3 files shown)


terminalforlife:
Ignore the Shellcheck nonsense here. No word-splitting or globbing occurs because it's a damn integer, and `-r` is not used to avoid expansion of escape sequences, and simply because it's not needed. Shellcheck should be taken with a grain of salt at times.

x70b1:
This project is alive. But there is no SLA for a pull request response.

In battery-cyberpower.sh you changed all variable names and otherwise only changed 2 quotes. Should this already be called optimization?

`shellcheck` is not always exact. But it helps to objectively improve the code. That is why we have it here. That is why we will continue to use it.

terminalforlife:
Yeah, I changed the variable names to the usual camel case, keeping them shorter and more consistent. As for optimization, that's mostly down to de-cat-ing and occasionally avoiding unnecessary uses of extra processes.

PR #216, opened 2020-03-04
vpn-networkmanager-status: VPN status command for NetworkManager/nmcli

sn0wvall:
When trying to use the vpn-openvpn-status script, I got simply got the message "nm-openvpn" - probably more a limitation of NetworkManager than the script. Therefore, I put together this script which will output the active NetworkManager-started VPN.


x70b1:
Just for understanding, does it not work with [network-networkmanager](https://github.com/polybar/polybar-scripts/tree/master/polybar-scripts/network-networkmanager)?
I guess we would have to add a new `network type`? This would give you the benefit of realtime updates.

sn0wvall:
The scripts are similar, however this script is more minimal providing only the VPN connection name, and excluding the redundant "tun0" from the output of network-networkmanager.

x70b1:
Nice work, thx!

x70b1:
I rewrote some parts. Maybe you want to check it: https://github.com/polybar/polybar-scripts/commit/0424be6f4c11a31cb56044239f52ab4fac7365bb

PR #223, opened 2020-03-31
Rewrite updates-fedora to work without sudo privileges

dannycolin:
Hi,

I've rewritten for my own use the updates-fedora module to work without adding `dnf` to `/etc/sudoers` and thought maybe that's something you'd be interested to merge in this repo :).

### DIFF
--- a/polybar-scripts/updates-fedora/README.md
+++ b/polybar-scripts/updates-fedora/README.md
@@ -4,10 +4,18 @@ A script that shows if there are updates for Fedora or `dnf` based distributions
 
 ![updates-fedora](screenshots/1.png)
 
-
 ## Configuration
 
-You have to add the `dnf` command to the `/etc/sudoers` NOPASSWD of your user:
+You can choose the `dnf` command used to retrieve the update information by 
+changing the number at the end of the `exec` line. Currently, there's two modes:
+
+1. updateinfo (default)
+   Shows uninstalled updates but doesn't show new dependencies
+2. update
+   Shows uninstalled updates and new dependencies
+
+For the second mode, you have to add the `dnf` command to the `/etc/sudoers` 
+NOPASSWD of your user:
 
 ```
 user ALL=(ALL) NOPASSWD: /usr/bin/dnf
@@ -18,6 +26,10 @@ user ALL=(ALL) NOPASSWD: /usr/bin/dnf
 ```ini
 [module/updates-fedora]
 type = custom/script
-exec = ~/polybar-scripts/updates-fedora.sh
+
+format = <label>
+label = %output%
+
+exec = ~/polybar-scripts/updates-fedora.sh 1
 interval = 600
 ```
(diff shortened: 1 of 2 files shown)


x70b1:
Thas a good benefit. 
But I am not sure why we did not used `updateinfo list`.
I guess a point was that this not refreshable. Could that be? I have no Fedora system here at the moment.

dannycolin:
> I guess a point was that this is not refreshable. Could that be? I have no Fedora system here at the moment.

The `--refresh` option can be used with all commands according to the dnf documentation. I think the only difference I can see between `update` and `updateinfo` is that if you have a update that needs a reboot, it'll continue to show as long as you don't reboot vs. `update` that only shows uninstalled packages. Personally, I like the former but I can understand that some users will prefer the latter.

If this is a problem, we could use `dnf list --upgrades` instead that shows the same list has `dnf update`. Or even better, I could add an option in my script that let the user choose which command they want to use. Something like:

```
update_mode=N

case $update_mode in
  1) 
    dnf update...
    ;;
  2)
    dnf updateinfo...
    ;;
  3)
    dnf list...
    ;;
esac
```

dannycolin:
So, I've rewrite the script to give the user the choice between the current `update` command or `updateinfo` that doesn't need `sudo`.

polybar-scripts/updates-fedora/updates-fedora.sh
```diff
@@ -1,14 +1,33 @@
...
+  2)
+    dnf=$(env LC_ALL=C sudo dnf upgrade --assumeno 2> /dev/null)
+    
+    upgrade=$(echo "$dnf" | grep '^Upgrade ' | awk '{ print $2 }')
+    install=$(echo "$dnf" | grep '^Install ' | awk '{ print $2 }')
+
+    updates=$(( upgrade + install ))
+    ;;
+  *)
+    updates=$(dnf -q updateinfo --list --updates 2> /dev/null | wc -l )
```
x70b1, line 25: I tested it on Fedora. `dnf -q updateinfo --list --updates 2> /dev/null` looks good. I am still not sure what our driver behind `dnf upgrade --assumeno` was. We should go with your solution, remove everything else to keep it clean and simple.

polybar-scripts/updates-fedora/README.md
```diff
@@ -18,6 +26,10 @@ user ALL=(ALL) NOPASSWD: /usr/bin/dnf
 ```ini
 [module/updates-fedora]
 type = custom/script
-exec = ~/polybar-scripts/updates-fedora.sh
+
+format = <label>
+label = %output%
```
x70b1, line 31: It's useless as its the default.

dannycolin:
> I played around with that a bit. A problem is that I got a different update count. I think that should be cleared out.

Strange. I'll investigate it this week. I'll keep you posted.

x70b1:
@dannycolin Any news here?

PR #228, opened 2020-04-13
Add package script that uses inotifywait.

jameh:
I ended up writing this script which uses inotifywait to update the status after pacman and yay transactions. It watches the directories `/var/lib/pacman.d/sync` and `"$XDG_CACHE_HOME/yay"` for file events and updates the package counts when `pacman` or `yay` is finished.

You can also poke it with a `USR1` signal to trigger an update, otherwise it fetches an update list every 15 minutes.


jameh:
I've realized that my forked process handling is not correct - so I'm closing this PR while I fix the problem

PR #230, opened 2020-04-16
vpn-networkmanager-wireguard: add script

sibizaestruch:
I saw there's was another script for Wireguard and based mine on that, this one uses `nmcli` and there's no need to do extra configuration.


x70b1:
What about [`vpn-networkmanager-status`](https://github.com/polybar/polybar-scripts/tree/master/polybar-scripts/vpn-networkmanager-status) or [`network-networkmanager`](https://github.com/polybar/polybar-scripts/tree/master/polybar-scripts/network-networkmanager) ?

sibizaestruch:
The script is more oriented to be used with wireguard interfaces, or even other interface, using just `nmcli` to do the job, without special configuration. 

Could be renamed as well, because can be used in any interface available in NetworkManager.

x70b1:
Does `nmcli -t connection show` print the wireguard connection as type = vpn? It seems not. It looks like the type is `wireguard`. 

In this case `vpn-networkmanager-status` should work,y we only have to add the `wireguard` type. The only difference is that there is no toggle at the moment. But that could be added. 

On the other hand should also `network-networkmanager` work. Its a bit bigger but it is realtime. The only thing is that there is not support for `connection-type = vpn` or `connection-type = wireguard`. But this is only a small problem. We could add this.

I can't see what should be wireguard specific in your script.

sibizaestruch:
You are right, the script can be used in any interface, not only for wireguard, the easy solution here can be adding the toggle to `vpn-networkmanager-status` and get the interface which can be `vpn` or `wireguard` type.

We can close this PR if you want, and I will send another one.

x70b1:
That would be cool.

PR #232, opened 2020-04-23
add docker container module

FalcoSuessgott:
This adds a docker-container module, which displays the number of container for certains status


polybar-scripts/docker-container/docker-container.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
```
x70b1, line 1: Use `#!/bin/sh` for posix compliance.

FalcoSuessgott: fixed with [#9e3634d](https://github.com/polybar/polybar-scripts/commit/9e3634d6bbc35f8f92321898d1daa85ef7ce5f76)

PR #234, opened 2020-04-27
Add pulseaudio-microphone

santiprandi:
Hi, I made a script that shows if the default microphone is muted or not. Useful for calls.

### DIFF
--- a/polybar-scripts/pulseaudio-microphone/pulseaudio-microphone.sh
+++ b/polybar-scripts/pulseaudio-microphone/pulseaudio-microphone.sh
@@ -1,0 +1,50 @@
+#!/bin/sh
+
+#
+# Prints the default microphone status (muted or unmuted). Uses PulseAudio.
+#
+
+# Global variables
+DISABLED_COLOR="#777" # Color when the mic is off.
+ENABLED_INDICATOR="#1" # Indicator when the mic is on.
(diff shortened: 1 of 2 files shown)


polybar-scripts/pulseaudio-microphone/pulseaudio-microphone.sh
```diff
@@ -0,0 +1,41 @@
...
+    LANG=EN; pactl subscribe | while read -r event; do
+        if echo "$event" | grep -q "source" || echo "$event" | grep -q "server"; then
+            status
+        fi
+    done
+}
+
+toggle() {
+  MUTED=$(pacmd list-sources | awk '/*/,EOF {print}' | awk '/muted/ {print $2; exit}')
+  DEFAULT_SOURCE=$(pacmd list-sources | awk '/*/,EOF {print $3; exit}')
```
x70b1, line 25: This line does not work here. Di you try to get the id? That would be the best.

santiprandi: Yes, it gets the id of the default source so it can use it to toggle the state. I think the problem was in the regex pattern not working in some distros because of the asterisk. It should be fixed now.

x70b1:
Thax for that! :sunglasses:

Issue #236, opened 2020-04-27
system-bluetooth-bluetoothctl: read: arg count


druckdev:
The dash (/bin/sh) read built-in expects a variable name and therefore fails in line 4 in the system-bluetooth-bluetoothctl script.
I was able to solve this by executing the script with /bin/bash instead of /bin/sh.

x70b1:
Discussion in #235.

Issue #235, opened 2020-04-27
system-bluetooth-bluetoothctl: piping commands instead of giving them as arguments


druckdev:
The system-bluetooth-bluetoothctl script was not working for me because my bluetoothctl version (I am running Ubuntu 18.04, bluetoothctl 5.48) does not accept commands as arguments. It will just start normally and won't exit by itself.
I was able to fix it by consulting https://wiki.archlinux.org/index.php/Bluetooth#Front-ends and changing the script so that the commands are being piped into bluetoothctl as suggested in the wiki.

x70b1:
Thanks for you PR! :sunglasses: 

Thats a bit a `wontfix`. The usage of `/bin/sh` ensures that no specific shell needs to be installed and used.

The most users will use the script with more up to date versions. Also newer versions of Ubuntu are out. 

If this works for you now, thats fine.

druckdev:
Thank you for your reply, that makes sense.

Could we specify a variable that is not used? By doing that /bin/sh can still be used, but the script would be more portable for the different versions.

Also what is with the piping into bluetoothctl?

x70b1:
I use `dash` to, but I dont see any problems here. 

Pipe the args to `bluetoothctl` works, but running the command as argument looks much cleaner and shorter. So I would leave it as it is.

PR #237, opened 2020-04-27
system-bluetooth-bluetoothctl: use bash and piping bluetoothctl commands

druckdev:
See https://github.com/polybar/polybar-scripts/issues/235 and https://github.com/polybar/polybar-scripts/issues/236


druckdev:
Thanks for your reply. I can understand that.

PR #251, opened 2020-06-07
Add ipfinder

muntjaber:
This script prints your public IP address (if available) and the status of your VPN connection. Since it makes use of two free (but limited) lookup services, it checks whether there has been actual changes to your network before querying for a new lookup. Accordingly, it responds quickly to network changes (within 2 seconds).


muntjaber:
Yes, please open a separate PR. 

Can you please elaborate on the issue of specificity? Currently, the script solves two issues: (1) what is my public IP address and (2) what is my VPN status. Do you mean I should separate them instead?

x70b1:
We cannot have every issue here as a scripted solution.
This repo is more a collection of ideas than a complete solution. 

We have already a script for `ifconfig.co` to find you the external IP. Also we have already 7 scripts which show the local VPN status.
I want to keep things simple. 

What we do not have yet is ipinfo.io
If you want to build a script out of that we could add it.

muntjaber:
Alright, thanks for the explanation. I'll create a script for `ipinfo.io` only and submit that instead.

x70b1:
Call it `network-ipfinder`.

muntjaber:
Sure! Thanks for the review. I'll close this pull request because of the aforementioned reasons.

PR #255, opened 2020-06-16
Fixes sunrise and sunset show time

peterhgombos:
The openweather fullfeatured script does not correctly show the sunrise and sunset now, as the time returned from the API is not actually shown to the user. Furthermore, the script converts the timezone to UTC. These commits makes sure that the correct sunrise and sunset is used, and that the current timezone is shown to the user.


x70b1:
Thx! I will look into this in the next days.

x70b1:
I can't reproduce the problem. What is your output and what is your expectation?

peterhgombos:
Oh, man. I really messed up. I thought the feature was to show the time of sunrise/sunset, and did not understand the output I got. 

Maybe the readme of the script should be changed from "It displays information about the next sunrise or sunset." to "It displays time to the next sunrise or sunset" to make it a bit clearer? :)

x70b1:
I improved it: b11a5a2efa9ff85d21db8369bafc264b4f28ab9a

PR #259, opened 2020-06-21
Add: Optimus-manager script for displaying active gpu and switching gpu.

OscarSanner:
Hello! I wrote a script that displays the active gpu in optimus laptops if the switching is managed by [optimus-manager](https://github.com/Askannz/optimus-manager). The script can also switch gpu (through optimus-manager) and is customizable to both "intel/nvidia" configurations and "hybrid/nvidia" configurations. I think it would make a fine addition to the collection.


polybar-scripts/info-optimus-manager/info-optimus-manager.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
```
x70b1, line 1: lets use `/bin/sh` for this

polybar-scripts/info-optimus-manager/README.md
```diff
@@ -0,0 +1,24 @@
...
+1. Make sure you have `enable-ipc=true` in the polybar config file.
+2. Edit the script and change the "Configuration" part to your liking. This is not necessary if you don't use the "Hybrid" mode in optimus-manager and you're happy with the default outputs. 
+3. (Optional, enables switching GPU when the module is left-clicked). Uncomment the line `;click-left = polybar-msg hook info-optimus-manager 2` in the module.
+
+
+## Module
+
+```ini
+[module/info-optimus-manager]
+type = custom/ipc
```
x70b1, line 18: Is there a reason why you use ipc and not a click event?

OscarSanner: Yes, originally I wanted to have an additional feature where if you clicked once the module would ask you to confirm, with a timer ticking to the left of the text. I thought it would be appropriate to use ipc for this. This was however scrapt. 

The reason I kept it was because it's unnecessary for Polybar to continually poll for data (i.e run the script at a given interval), since the module is for a powersaving feature in laptops and since the gpu only ever changes in between sessions. 

However, it's indeed probably better to just use the common `custom/script` type and set it to execute at a rather high interval. Readme will be updated with the next commit.

polybar-scripts/info-optimus-manager/info-optimus-manager.sh
OscarSanner: this has been fixed in a later commit and all the bash specific language has been removed :)

polybar-scripts/info-optimus-manager/info-optimus-manager.sh
```diff
@@ -0,0 +1,85 @@
...
+	elif [ "$(get_current)" = "nvidia" ]; then
+		if [ "$hybrid_switching" = 1 ]; then
+			next="hybrid"
+		else
+			next="intel"
+		fi
+	elif [ "$(get_current)" = "hybrid" ]; then
+		next="nvidia"
+	fi	
+	eval "optimus-manager --switch $next --no-confirm"
```
x70b1, line 69: Is here a reason for eval?
You could just run the command?

polybar-scripts/info-optimus-manager/README.md
```diff
@@ -0,0 +1,23 @@
+# Script: info-optimus-manager
```
x70b1, line 1: I think we should rename this script and the directories to `system-gpu-optimus`.

x70b1:
Awesome, thx! :sunglasses:

x70b1:
I updated the script to bring it a bit in shape. 
Can you test it?

PR #260, opened 2020-06-30
player-cmus: handle no file information

mpriestman:
When CMUS is started for the first time, cmus-remote doesn't return
any track data, which means $duration is an empty string. This causes
the script to fail with an error on line 12 when it tries to evaluate
$duration as an integer.

The fix is to only print any information if we have either a title or
an artist, which means we probably also have a duration as well.

### DIFF
--- a/polybar-scripts/player-cmus/player-cmus.sh
+++ b/polybar-scripts/player-cmus/player-cmus.sh
@@ -2,33 +2,37 @@
 
 if info=$(cmus-remote -Q 2> /dev/null); then
 	status=$(echo "$info" | grep -v "set " | grep -v "tag " | grep "status " | cut -d ' ' -f 2)
-	
+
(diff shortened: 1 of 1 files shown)


polybar-scripts/player-cmus/player-cmus.sh
```diff
@@ -2,33 +2,37 @@
...
+			elif [ "$status" = "paused" ]; then
+				echo "#2 $info_string"
+			elif [ "$status" = "stopped" ]; then
+				echo "#3 $info_string"
+			else
+				echo ""
+			fi
 		else
-			echo ""
+			echo "";
```
x70b1, line 35: Was the `;` intended?

mpriestman: Oh, no, that's probably a mistake. Let me fix that...

PR #262, opened 2020-07-09
Add my Toggl.com time tracker widget to the list

maksmeshkov:
It's a simple widget that uses toggl.com API to get your current running task info and display it.


x70b1:
Thx! :sunglasses:

PR #265, opened 2020-07-30
system-fan-speed: fix empty speed

muntjaber:
Fix issue with speed expression returning empty string instead of integer.

### DIFF
--- a/polybar-scripts/system-fan-speed/system-fan-speed.sh
+++ b/polybar-scripts/system-fan-speed/system-fan-speed.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-speed=$(sensors | grep fan1 | cut -d " " -f 9)
+speed=$(sensors | grep fan1 | awk '{print $2; exit}')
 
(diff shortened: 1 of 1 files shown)


x70b1 (commented):
Sorry for the delay. 

I guess it would also a good idea to use the condition in line 5. 
Does something like  `if [ -n "$speed" ]; then` work?

muntjaber:
Yes, that is required. If `fan1` doesn't exist, an empty string will be returned which results in a syntax error when evaluating `speed_round`.

x70b1:
Yes. But is a real problem?

Are threre cases where your system rename a fan?

muntjaber:
No, not that I am aware of. However, using the current solution, it failed to identify the field using a count of 9 spaces (returning an empty string). Hence, the proposed changed: delimit by any number of spaces instead, and select the second field.

x70b1 (approved):

x70b1:
Thx! :sunglasses:

PR #266, opened 2020-07-31
Add info-active-window

andloh:


x70b1 (commented):
Your script terminates after one run. The `tail = true` can not work here.

PR #268, opened 2020-09-06
Dunst status

elshize:
Notification script for `dunst` using `dunstctl is-paused`. See README for more information.


polybar-scripts/dunst-status/README.md
```diff
@@ -0,0 +1,53 @@
+# Dunst Status & Toggle
+
+This is a simple script that displays dunst's `is-paused` status.
+It will display a crossed out bell icon
+if the notifications are paused,
+or a regular bell otherwise.
+
+![](screenshots/on.png)
```
x70b1, line 8: Lets rename the screenshots to `1.png` and `2.png`.

polybar-scripts/dunst-status/dunst-status.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/sh
+
+ON=""
```
x70b1, line 3: Replace them with `#` as described in the contributions.md.

polybar-scripts/dunst-status/README.md
```diff
@@ -0,0 +1,53 @@
+# Dunst Status & Toggle
```
x70b1, line 1: We should rename that script and all paths to `info-dunst`.

polybar-scripts/dunst-status/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Module Configuration
+
+```
+[module/dunst]
+type = custom/script
+format = <label>
+label = %output%
+exec = /path/to/dunst-status.sh
+tail = true
```
x70b1, line 31: Why do you use the `tail = true` ?
You could also run that script with an interval of 1. But that will use a lot of resources. A higher interval would be better.

x70b1 (changes requested):

polybar-scripts/dunst-status/README.md
elshize: The reason for that is to have an immediate response once the status is changed. This is similar to the example in the polybar repo: https://github.com/polybar/polybar/wiki/Module:-script#examples

If you have just a one-shot script on an interval, the change to the status won't happen right away, only after the interval. Which might seem like it's not working or simply be annoying, especially with higher interval. That's why tail is `true`, and clicking on the status icon sends a signal to the process to update. As explained below in the README, it also allows for having a script that toggles the status from, say, a key shortcut. This way, as long as `dunstctl` isn't called manually, we have a seamless experience. If it is, then it still works, but with that up-to-1-second lag. I can add a comment suggesting increasing the interval in the script if one doesn't plan to use `dunstctl` manually to change the status.

PR #274, opened 2020-10-10
notification-xkcd: minor fixes

kippenjongen:
This pull request adds handling for if the save file doesn't exist (now it creates a new one, previously it crashed), and removes unnecessary writes to the save file.


x70b1 (approved):

x70b1:
Thanks! :smiley:

PR #276, opened 2020-10-18
network-traffic: fix incorrect speed calculation math

MicroDroid:


MicroDroid:
this is wrong actually

PR #278, opened 2020-11-02
i3 layout indicator

rddunphy:
This is a Python script that runs in the background and updates a Polybar indicator to show the layout mode for the currently focused window in the i3 window manager. (It's intended mostly for users who disable the rather ugly window frame indicator.) It subscribes to key bindings that change the layout, window focus events, and tick events (tick events allow the user to manually update the indicator if they change the layout by some method that doesn't generate an i3 event). It also handles the special case of empty workspaces, so that the indicator always shows where the next window will be placed.

The script updates a `custom/ipc` module named `layout`. The default indicator text in the module config example is "V" for vertical split, "H" for horizontal split, "T" for tabbed, and "S" for stacked. (There's also a fifth hook included as a fallback for any other layout mode - I can't find any situation that would trigger this at the moment, so this is mainly to help with debugging in case i3wm's API changes at some point in the future.)

Aside from Python 3 and i3, the script has one requirement - the [i3ipc](https://pypi.org/project/i3ipc/) Python library, which can be installed using `pip`.


x70b1:
Thanks for this PR @rddunphy!

PR #280, opened 2020-11-07
player-mpris-tail: fixed blacklist bug in onChangedProperties

chenxu2048:
I run player-mpris-tail.py and add chrome into blacklist. However, the module output changed while chrome player status changed. Although chrome is in the blacklist, it does not check in  ``onChangedProperties``, and chrome will be recognized as the new player.


tfaughnan:
I ran into this same issue with blacklisting. Because there is currently no check in `onChangedProperties`, `addPlayer` is called with the argument `bus_name=None` which creates a Player object for the blacklisted player. This PR from @chenxu2048 solves it and should be merged.

x70b1:
Thanks for that feedback!

x70b1 (approved):

PR #282, opened 2020-11-11
updated the fontawesome5 icons

CedBri:
Using FontAwesome5 the icons displayed sound icons (play, pause, stop) instead of weather icons. Updated it using their [cheatsheet](https://fontawesome.com/cheatsheet).


x70b1:
You used FontAwesome 5 Free not Pro. 
Thats why you have different icons.

PR #283, opened 2020-11-13
Allow change of mic volume using the scroll wheel

radiantly:
The module also shows the current volume percent (when not muted)

Cheers!

### DIFF
--- a/polybar-scripts/pulseaudio-microphone/pulseaudio-microphone.sh
+++ b/polybar-scripts/pulseaudio-microphone/pulseaudio-microphone.sh
@@ -4,9 +4,9 @@ status() {
   MUTED=$(pacmd list-sources | awk '/\*/,EOF {print}' | awk '/muted/ {print $2; exit}')
 
   if [ "$MUTED" = "yes" ]; then
-    echo "#1"
+    echo "muted"
   else
-    echo "#2"
+    pacmd list-sources | grep "\* index:" -A 7 | grep volume | awk -F/ '{print $2}' | tr -d ' '  
   fi
 }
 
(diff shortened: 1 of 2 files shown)


polybar-scripts/pulseaudio-microphone/pulseaudio-microphone.sh
```diff
@@ -25,16 +25,32 @@ toggle() {
   DEFAULT_SOURCE=$(pacmd list-sources | awk '/\*/,EOF {print $3; exit}')
 
   if [ "$MUTED" = "yes" ]; then
-      pacmd set-source-mute "$DEFAULT_SOURCE" 0
+      pactl set-source-mute "$DEFAULT_SOURCE" 0
   else
-      pacmd set-source-mute "$DEFAULT_SOURCE" 1
+      pactl set-source-mute "$DEFAULT_SOURCE" 1
```
x70b1, line 30: Why du you use pactl instead of pacmd? 

As pacmd is already part of the pulseaudio package, pacmd would be a new dependency. So we could use it, but to have fewer requirements is also a nice goal.

radiantly: I believe the original reason for using `pactl` instead of `pacmd` is that it does not allow you to increase the volume by a certain percent - `pactl set-source-volume "$DEFAULT_SOURCE" +5%`

x70b1: Thats what I had in my head too. Alright then leets keep it like this.

x70b1 (approved):

PR #294, opened 2020-12-21
Add stopwatch

pryme-svg:


sorinraduta (approved):

x70b1:
Looks cool.

PR #297, opened 2021-01-04
player-mpris-tail: Handle empty array metadata correctly. Fixes #295

### LINKED ISSUE #295: Issue with mpris-tail and ncspot before playback starts
I'm using mpris-tail with [ncspot](https://github.com/hrkfdn/ncspot). When the player first opens, but before playback starts, the script produces the following error: 

~~~
Traceback (most recent call last):
  File "/home/andrew/dotfiles/scripts/polybar/mpris-tail", li
ne 502, in <module>
    PlayerManager(blacklist = args.blacklist)
  File "/home/andrew/dotfiles/scripts/polybar/mpris-tail", li
ne 32, in __init__
    self.refreshPlayerList()
  File "/home/andrew/dotfiles/scripts/polybar/mpris-tail", li
ne 86, in refreshPlayerList
    self.addPlayer(player_bus_name)
  File "/home/andrew/dotfiles/scripts/polybar/mpris-tail", li
ne 92, in addPlayer
    player = Player(self._session_bus, bus_name, owner = owne
r, connect = self._connect, _print = self.print)
  File "/home/andrew/dotfiles/scripts/polybar/mpris-tail", li
ne 198, in __init__
    self.refreshMetadata()
  File "/home/andrew/dotfiles/scripts/polybar/mpris-tail", li
ne 248, in refreshMetadata
    self._parseMetadata()
  File "/home/andrew/dotfiles/scripts/polybar/mpris-tail", li
ne 284, in _parseMetadata
    self.metadata['artist']     = re.sub(SAFE_TAG_REGEX, """\
1\1""", _firstIfList(_artist))
  File "/usr/lib/python3.9/re.py", line 210, in sub
    return _compile(pattern, flags).sub(repl, string, count)
TypeError: expected string or bytes-like object
~~~

It's happening because at [L284](https://github.com/polybar/polybar-scripts/blob/master/polybar-scripts/player-mpris-tail/player-mpris-tail.py#L284) `_artist` is set to `[]` by ncspot and `_firstIfList`, because that method checks the length of any list passed, returns `[]`, which doesn't work.

Changing L284 to `self.metadata['artist']     = re.sub(SAFE_TAG_REGEX, """\1\1""", _firstIfList('' if _artist == [] else _artist))` makes the script work again, but this feels like a hack (otherwise I'd submit a PR).

I'm assuming this error is happening because something isn't working in ncspot, but I wanted to post it here, in case this is a broader issue, also there is a problem that `firstIfList` can still return a list in certain cases.

rockyprabowo:
This PR will fix an issue when metadata(s) with empty array value `[]` is sent by the player.

Also, this commit will rename the function `_firstIfList` to `_metadataGetFirstItem`. The new name reflects to what the function actually do.

I'm sorry for the messy PR I did before. I forgot how to GitHub.

### DIFF
--- a/polybar-scripts/player-mpris-tail/player-mpris-tail.py
+++ b/polybar-scripts/player-mpris-tail/player-mpris-tail.py
@@ -281,18 +281,18 @@ class Player:
             _cover      = _getProperty(self._metadata, 'xesam:artUrl', '') or _getProperty(self._metadata, 'mpris:artUrl', '')
             _duration   = _getDuration(_length_int)
             # Update metadata
-            self.metadata['artist']     = re.sub(SAFE_TAG_REGEX, """\1\1""", _firstIfList(_artist))
-            self.metadata['album']      = re.sub(SAFE_TAG_REGEX, """\1\1""", _firstIfList(_album))
-            self.metadata['title']      = re.sub(SAFE_TAG_REGEX, """\1\1""", _firstIfList(_title))
+            self.metadata['artist']     = re.sub(SAFE_TAG_REGEX, """\1\1""", _metadataGetFirstItem(_artist))
+            self.metadata['album']      = re.sub(SAFE_TAG_REGEX, """\1\1""", _metadataGetFirstItem(_album))
+            self.metadata['title']      = re.sub(SAFE_TAG_REGEX, """\1\1""", _metadataGetFirstItem(_title))
             self.metadata['track']      = _track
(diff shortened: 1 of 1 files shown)


x70b1:
@rockyprabowo Thanks for that PR and working on that issue! 
Sorry for the delay. 

@Cybolic 
What do you think?

Cybolic:
Thanks for the PR, looks good to me!

x70b1 (approved):

PR #298, opened 2021-01-07
cpu-frequency: add script

muntjaber:


x70b1:
We have already a script for this: [system-cpu-frequency](https://github.com/polybar/polybar-scripts/tree/master/polybar-scripts/system-cpu-frequency)

muntjaber:
Seems like I missed that. Thanks for pointing that out!

I'll close this request accordingly.

PR #299, opened 2021-01-07
Added info on how to use own fonts

Kuuhhl:
I wanted to use my font (that I use for the rest of my polybar) but still keep the weather icon. 
I spent quite a lot of time trying to find out how to change it so I figured I would make it easier and just add it to the README (since changing the font to match your main one is something most of the script's users will want to do.)


x70b1:
Congrats! You found out how it works. :sunglasses: 

You could even be more explicit with `%{T1}` or T2 .. :

```
echo "%{T3}""$(get_icon "$weather_icon")" "%{T-}""%{T1}$weather_temp%{T-}$SYMBOL
```

I know the fonts are a bit tricky. But I think that's part of the learning curve.

PR #301, opened 2021-01-09
add info-docker-advanceadd info-docker-advance

stiliajohny:
info-docker-advance works with rofi to show a menu when you left-click or right-click on the polybar module

Users can select the running container or select existing images
for the running containers, the user can select from bringing up a TTY, Stop, Pause, Unpause, remove and Log tail the container


stiliajohny:
Hi 
anyone is looking at this PR ?

stiliajohny:
Perhaps @x70b1 ??

stiliajohny:
hi is anyone is looking on this PR ?

x70b1:
Hi @stiliajohny 

As a simple example we have the `info-docker` script.

With the rofi part it is a bit out of scope for this repo I would say. 
Also as docker will become less important in the future I don't want invest to much time here.

But if you want to setup an own reposiItory for this project, I would add it to the repo list on the main README.md.

stiliajohny:
Will set u pa repo as I have quiet few of modules. I will followup on this

PR #303, opened 2021-01-15
[Changed README.md] Added left-click option to open nmtui

BreenBrain:
Added left-click option to the module itself. Now when the module is clicked it will open nmtui.


o-dka:
Bruh

x70b1:
Well, I dont see a bigger improvement of the script here. 
Adding `nmtui` can be helpful, but I think it is more a personal use case.

PR #304, opened 2021-01-16
added coronavirus module

Kuuhhl:
This plugin adds a widget to polybar that shows the current number of infections in the specified area.
I also implemented a cache-file to be able to compare numbers and add an arrow. It is also used if the the internet doesn't work. 
I tried adding command-line arguments for most things to make the script easier to customize.


x70b1:
Done.

PR #308, opened 2021-01-29
system-bluetooth-bluetoothctl: Remove pipe

MTBorg:
Replaces a pipe into a "`while read`"-statement with a simple for loop.
This has the advantage that a new subshell is not created because of the pipe, meaning the counter variable can be used outside of the loop. Also, i'd argue it's easier to read.


x70b1 (approved):
Thats right. It works as long as there is no space in the device.
But that should not be the case here.

Good catch!

PR #311, opened 2021-02-04
Add info-nicehash module

theblazehen:


x70b1 (commented):
Thanks for this PR!

PR #313, opened 2021-02-22
add player-mpris-scroll

xxyangyoulin:


x70b1:
Thanks for this PR @xxyangyoulin 

As we have already `player-mpris-tail` based on python and `player-mpris-simple` based on playerctl I dont warn to pull this script.

PR #317, opened 2021-03-29
mic-volume script

AggamR:
I made a mic-volume script to show the volume of the mic, and a polybag module for it (obviously).


AggamR:
Found a more efficient solution, will make a req

PR #318, opened 2021-03-29
mic-vol module

AggamR:
I've made this module to see the volume of the mic. no need for an external shell script as well!


x70b1:
Well its an one liner. That is too little to keep it here.

PR #321, opened 2021-04-26
Added support for multi monitor setups to herbstluftwm workspace info

fanginoort:
This adds support for using the info-hlwm-workspaces.sh script on a multi monitor setup.

It expects MONITOR environment variable to be set for each instance of polybar. If the MONITOR variable is not set this will still be working, however there will be no distinction between the different monitors.
This enables formatting the tags considering whether the current monitor is focused or not.

Also if one clicks on one of the tags in polybar, the script will ensure that the monitor is focused where the user clicked the tag. This prevents the effect that if one clicks a tag on an unfocused monitor, the tag will change on the other, focused monitor. This is eventually unexpected behaviour.

### DIFF
--- a/polybar-scripts/info-hlwm-workspaces/info-hlwm-workspaces.sh
+++ b/polybar-scripts/info-hlwm-workspaces/info-hlwm-workspaces.sh
@@ -18,26 +18,33 @@ herbstclient --idle "tag_*" 2>/dev/null | {
             for i in "${tags[@]}" ; do
                 # Read the prefix from each tag and render them according to that prefix
                 case ${i:0:1} in
-                    '#')
-                        # the tag is viewed on the bar's monitor and is focused
-                        # TODO Add your formatting tags for focused workspaces
-                        ;;
-                    '+')
-                        # the tag is viewed on the bar's monitor, but is not focused
-                        # TODO Add your formatting tags for viewed but unfocused tags
-                        ;;
-                    '!')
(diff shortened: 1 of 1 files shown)


polybar-scripts/info-hlwm-workspaces/info-hlwm-workspaces.sh
```diff
@@ -1,40 +1,51 @@
...
         {
             for i in "${tags[@]}" ; do
                 # Read the prefix from each tag and render them according to that prefix
                 case ${i:0:1} in
                     '#')
-                        # the tag is viewed on the focused monitor
+                        # the tag is viewed on the bar's monitor and is focused
                         # TODO Add your formatting tags for focused workspaces
                         ;;
-                    ':')
```
x70b1, line 16: Why do you replace the existing cases instead of creating new ones?

patrick96 (changes requested):
It would probably be a good idea to have a case for each state prefix instead of replacing some of the cases from before.

Otherwise this looks good :)

patrick96:
Seems x70b1 was slightly faster :wink:

fanginoort:
I tried to change as little as possible, which however wasn't the smartest thing after all.
The new commit now includes one case for each state prefix with the comments comming right from herbstluftwm's documentation :)

patrick96 (approved):
LGTM!

x70b1:
Thanks @fanginoort !

PR #326, opened 2021-05-12
Add Info-Podman script

RealStickman:
This PR adds a script similar to info-docker for podman.  
This script assumes that podman is being used in rootless mode (as explained in the README).


x70b1 (approved):
Nice, thanks!

PR #327, opened 2021-05-21
Add formatted length tag for player-mpris-tail

dzshn:
Small addition I thought it'd be nice to have
adds another format tag for length that shows proper m:ss / h:mm:ss time format, instead of microseconds

Awesome script, by the way! I had made one myself earlier but ended up switching to this one, it's way faster


x70b1:
@Cybolic did a great job there. :sunglasses: 
Maybe he has some thoughts on this PR.

dzshn:
oops, I'm not much familiar with more advanced git workflow, is that an issue? I intended just a small change for it to use an hyphen instead of an underline as "icon-reverse" was like so..
I guess this PR could be closed and redone later when there's a response from him if necessary

x70b1:
Everything is fine :) We can keep it open. 
If there is a response we can work on it.

Cybolic:
I haven't tested this PR, but it looks okay to me :)
It's infinitely more useful than the current time format, so good job @dzshn!

PR #328, opened 2021-05-21
Fixed shebang for updates-pacman-aurhelper

druskus20:
The shebang in _updates-pacman-aurhelper_ is incorrect as it assumes that the user has sh symlinked to bash. That is the case by default in arch, but its still not a safe assumption.

Specifically, the line that is using a bash'ism is the following:
```sh
updates=$(("$updates_arch" + "$updates_aur"))
```

### DIFF
--- a/polybar-scripts/updates-pacman-aurhelper/updates-pacman-aurhelper.sh
+++ b/polybar-scripts/updates-pacman-aurhelper/updates-pacman-aurhelper.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 if ! updates_arch=$(checkupdates 2> /dev/null | wc -l ); then
     updates_arch=0
--- a/polybar-scripts/updates-pacman-aurhelper/updates-pacman-aurhelper.sh
+++ b/polybar-scripts/updates-pacman-aurhelper/updates-pacman-aurhelper.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 if ! updates_arch=$(checkupdates 2> /dev/null | wc -l ); then
     updates_arch=0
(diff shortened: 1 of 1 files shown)


x70b1 (commented):
Thanks for this PR!

We don't asume a bash. `/bin/sh` can be linked to every shell - zsh, bash, dash ..
Calculations with `$(( ))` are not bash-specific.

You can run this script with different shell and it will work. 
Asuming a specific shell would be a step back as it would be a hard dependency.

dzshn:
I just did some testing, on zsh it errors out with `updates-pacman-aurhelper.sh:16: bad math expression: illegal character: "`, so I'm guessing the error is having quoted variables inside the `$((  ))`
I think it's safe to depend on bash, as it's pretty much used as a dependency on a lot of stuff (for me there are 85 packages installed that depend on it), including really common software such as man-db, nano, mkinitcpio, gzip, p7zip, lvm2, automake etc. so it's actually really hard to have a system *without* bash.
On the other hand though, I'm not exactly sure what the quotes change within that expression, but that's because I'm more used to scripting for zsh, perhaps that could be replaced with another one that has equivalent behavior?

anywho, just my thoughts on this PR

x70b1:
The quotes are a problem here, thats right. 
So we need something like `updates=$((updates_arch + updates_aur))`.

druskus20:
zsh is not a good example, as its not 100% POSIX compatible. I myself am using dash and I also get the error `updates-pacman-aurhelper.sh:16: bad math expression: illegal character: `. 

@x70b1 suggestion is exacly what I am doing in my own -fixed- version of this script.

x70b1:
Yes, so I would prefer to fix the quotes, not the shebang.

druskus20:
Note that I've not been able to test this propperly, as I dont have any update pending right now

x70b1 (approved):
Yes, this works for bash, dash and zsh. 

Thanks @druskus20 and @dzshn :sunglasses:

PR #329, opened 2021-05-24
isrunning-noisetorch: add script

druskus20:
Added a script to show an icon in case that noisetorch is loaded as a pulseaudio module


x70b1 (approved):
Thanks for this PR. Thats cool!
I added some smaller fixes.

PR #330, opened 2021-06-02
Add Dunst notification switch

RealStickman:
This module display the current status of Dunst (enabled or disabled) and allows you to toggle that with a left-click.  

Sidenote: Could you please add the target shell to CONTRIBUTING.md? I'm not sure if you're aiming for POSIX sh compatibility, or if targeting Bash is fine.  
It might also reduce the amount of times you need to change the shebang and fix scripts if this is documented.


x70b1:
Thanks!

I have added a not to the CONTRIBUTING.md.

PR #331, opened 2021-06-13
added salah-time script

robinsrk:


dzshn:
Sorry if this is a bold statement, but what's with `checkDate.py`? Using pickle for storing what seems to be a dict isn't the most safe or appropiate - as the docs even point out:
> It is possible to construct malicious pickle data which will execute arbitrary code during unpickling. Never unpickle data that could have come from an untrusted source, or that could have been tampered with.

I would recommend using json for it..

robinsrk:
> Sorry if this is a bold statement, but what's with `checkDate.py`? Using pickle for storing what seems to be a dict isn't the most safe or appropiate - as the docs even point out:
> 
> > It is possible to construct malicious pickle data which will execute arbitrary code during unpickling. Never unpickle data that could have come from an untrusted source, or that could have been tampered with.
> 
> I would recommend using json for it..

it is storing the data, so that if internet is not working, or not available, it won't call the api again until next day. So the point of using that file is to store data for one day, and next day check the api and store the data.

robinsrk:
> Sorry if this is a bold statement, but what's with `checkDate.py`? Using pickle for storing what seems to be a dict isn't the most safe or appropiate - as the docs even point out:
> 
> > It is possible to construct malicious pickle data which will execute arbitrary code during unpickling. Never unpickle data that could have come from an untrusted source, or that could have been tampered with.
> 
> I would recommend using json for it..

Ok I will push new updates soon with json formatted data

x70b1:
Thanks for this PR!

PR #338, opened 2021-07-17
added mapping for Nerd Fonts and other changes

rustbuckett:
Added mapping for Nerd Fonts and extended the mapping to include more icons for other data. Commented forecast code and changed output line to include more of the data returned from openweathermap.org. Mappings for Weather Icons and Font Awesome would still need to be implemented.


x70b1:
Thanks for this!

Thats are probably a lot of changes at the same time.
But if you remove the changes with the forecast part and just add Nerd Font icons I would pull it.

PR #340, opened 2021-08-01
added scrolling to always have a fixed length text but still show the whole Title/Artist

Laubersheini:


x70b1:
Maybe @Cybolic wants to add some comments here.

PR #341, opened 2021-08-09
New script: bspwm window titles

kndndrj:
This script is used to display window titles for bspwm, all other info should be in the README.md


x70b1:
Sorry fo the delay. Thanks for this PR!

PR #344, opened 2021-08-24
system-cpu-avgcoretemp: This script displays the average temperature across all cores of the CPU in your system.

m-GDEV:
Hi there! I've decided to add one of my personal scripts to this amazing project you have here. 

I understand there is also a script called "system-cpu-temppercore" that is similar in function. Again just to make the distinction, this script display a whole number of the avergage temperature of all core in the system while the above mentioned plugin shows the individual temperatures of each core.

Thanks for the work you've done on this project! If you have any questions, comments, or concerns do not hesitate to contact me (email in bio).

Have a great day.


x70b1:
Sorry fo the delay. Thanks for this PR!

I only see a limited use case here for other users. But if you want to setup an own repository for this script, I would add it to the repo list on the main README.md in `polybar-scripts`.

PR #345, opened 2021-09-06
Add xfce4-presentation-mode script

cesalazar:
Toggles and displays xfce4-power-manager presentation mode.


x70b1:
Thanks for this PR!

PR #347, opened 2021-09-14
script to show the most memory consuming process

michael105:
Optionally kill the process on rightclick.


x70b1:
Thanks for this PR!

michael105:
Well, it is useful when doing development work.
As soon, e.g. a daemon starts to go crazy and keeps using more memory until the system blocks, this will be shown in the statusbar.
Same with my internet browser here.
I also write a script to show the most CPU consuming process for the same reason.

PR #348, opened 2021-09-26
Disk storage monitoring bash script

makccr:
I created a little module for my polybar to print the percent of storage used on the main system drive , / 

Thought it might be useful for some other users.


x70b1:
Thanks for this PR.
You can use the [existing module](https://github.com/polybar/polybar/wiki/Module:-filesystem) of polybar.

PR #353, opened 2021-11-25
hlwm color tags

RaZ0rr-Two:


x70b1:
Thanks for your work!
Maybe its a good idea to change the existing script and add your color tags there?

RaZ0rr-Two:
> Thanks for your work! Maybe its a better idea to change the existing script and add your color tags there instead of adding a second script?

I thought about it but I thought it might be better to leave both scripts so that newbies like myself can take a look at both and figure out much more on the customizations.

x70b1:
Your changes add a nice feature to the script, and the only thing why theese features are not there is because the original submitter had no time or use for it. So I see tham as a nice addition.

We are all here to learn things! :wink:

PR #358, opened 2021-12-28
Runcat Polybar

KaboomPhysicist:
This is a module for Polybar. It's a port from Runcat Extension for Gnome. It just shows a cat animated in the bar, whose speed depends on the cpu usage.


polybar-scripts/system-cpu-percentage-runcat/README.md
```diff
@@ -0,0 +1,32 @@
+# Script: system-cpu-percentage-runcat
+
+A python script which displays CPU usage and an animated cat according to the percentage.
+
+![runcat](screenshots/0.png)
```
x70b1, line 5: As the pictures are similar, I would suggest to use only one.

polybar-scripts/system-cpu-percentage-runcat/README.md
```diff
@@ -0,0 +1,32 @@
+# Script: system-cpu-percentage-runcat
```
x70b1, line 1: Lets call the directory and all files `system-cpu-runcat`.

polybar-scripts/system-cpu-percentage-runcat/speedcpu.py
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/python3
+
```
x70b1, line 2: Why do we have 2 files? On e file would be better?

polybar-scripts/system-cpu-percentage-runcat/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+You must add the new font to your system and include it in the Polybar config file. The font includes the icons for the cat. Just add the .ttf file to /usr/share/fonts and run fc-cache in shell.
+
+## Module
+
+```ini
+[module/runcat]
+type = custom/script
+exec = ~/polybar-scripts/system-cpu-runcat.py
+interval = 0.2
```
x70b1, line 30: Maybe `tail=true` would be good idea. The script would need to changed then.

x70b1 (changes requested):

x70b1:
Cool idea! And sorry for the delay.

Issue #314, opened 2021-03-02
Softwarecounter: Not displaying the rxvt logo


HugoGDRM:
Hi, i'm struggling to display the rxvt logo, I added 'urxvt' and 'rxvt' to the 'terminals' combine_guis but nothing happen. I checked with htop to determine the name and it is rxvt for sure.
Thnks

HugoGDRM:
However, chromium and vim logos are displayed perfectly

HugoGDRM:
urxvt-unicode doesn't works

x70b1:
Sorry for the late response. I can confirm the issue. `urxvt` seems not to be detected as terminal. 
If I put it as vim command to the list it works. But thats no stable solution.

@jtambasco contributed it. Maybe he has an idea.

PR #359, opened 2022-01-07
info-softwarecounter: the whole script has been refactored

jfernandz:
I have some improvments for this script I like very much. 

I hope you appreciate them. :wink:

BR and happy new year :blush:

### DIFF
--- a/polybar-scripts/info-softwarecounter/info-softwarecounter.py
+++ b/polybar-scripts/info-softwarecounter/info-softwarecounter.py
@@ -1,8 +1,8 @@
 #!/usr/bin/env python3
 
 import subprocess
-import sys
 from collections import Counter, OrderedDict
+import logging
 
 
 # update GUIs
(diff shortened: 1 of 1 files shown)


x70b1:
Awesome!
Is issue https://github.com/polybar/polybar-scripts/issues/314 also fixed then?

jfernandz:
hmmm, no, apparently it's not solved, but I didn't know about that issue, I'm also realizeing I could also include support for telegram :thinking: or spotify ...

I think I'll work on these :wink:

x70b1:
I looked into that issue some time ago but was not able to find the reason there. 
So if you would find a reason and a fix for this it would be very cool!

jfernandz:
Sure I'm checking it :wink:

jfernandz:
I'd say right know should work also for `urxvt` and `nautilus` :wink:

jfernandz:
I'm still trying to improve a few things about new logging feature. I'll tell you when it's done :blush:

PR #360, opened 2022-01-10
Improved system-ati-radeontop

Peter-Crawley:
After messing around with this a bit I found that I can make use of polybar's ability to listen for continuous output, which radeontop and grep are both capable of providing.

This newer configuration results in the same output without constantly starting new processes.


x70b1 (commented):
Then we could remove the interval parameter.

You removed sudo, is this not required? I dont have a system to check this.

Peter-Crawley:
On my system the sudo does not seem to be required, and I have not done anything that would make this be the case, so it should be fine.

x70b1 (approved):
Thanks for your work!

PR #361, opened 2022-01-18
bspwm-node-flags: script, module & preview

aidenlangley:


aidenlangley:
Could somebody translate this to sh/bash so I can include a not so fishy version? :)

```sh
set name (string split ':' $flag)[1]
```

x70b1:
> Could somebody translate this to sh/bash so I can include a not so fishy version? :)

Are you looking for `| cut -d ':' - f 1` ?

aidenlangley:
> > Could somebody translate this to sh/bash so I can include a not so fishy version? :)
> 
> Are you looking for `| cut -d ':' - f 1` ?
> 
> 

Awesome, yes ty. I've been spoiled by fish, the traditional stuff is not so easy. I also have to translate the `string match` and the iterator that follows then this one will be ready.

aidenlangley:
So I rewrote it in `python` for wider shell compatibility since I'm sure `python` is a lot more common than `fish`. Much less painful to write than pure `sh`, so I'm afraid that's the best I can do!

x70b1:
No pressure. 

Is there a way to replace the `bspwn-node-flags` script also with python? Or even better, combine them into one?

aidenlangley:
> No pressure.
> 
> Is there a way to replace the `bspwn-node-flags` script also with python? Or even better, combine them into one?

Way ahead of you. :D

I agree, python was a better choice here, have included a sh script which is intended to be tailed. Could have moved the `if arg.event == 'node_focus'` stuff to the `sh` script in hindsight for a slight enhancement but I'll leave it at that. :)

PR #362, opened 2022-01-18
dunstctl: script, module & preview images

aidenlangley:


aidenlangley:
Sure, can include a bash version, will push it shortly.

polybar-scripts/dunst-controller/dunstctl.sh
```diff
@@ -0,0 +1,50 @@
...
+    sleep 2
+    # Close all to prevent notification from showing up once we unmute.
+    dunstctl close-all
+    dunstctl set-paused true
+}
+
+pop_all() {
+    for id in $(dunstctl history | while jq -r '.data[][].id.data'); do
+        dunstctl history-pop $id
+    done
```
emanuele6, line 22: Are you sure this works? Did you forget to commit a change? This bit of code is not valid.

Here is how it should be

```suggestion
    dunstctl history | jq '.data[][].id.data' | while IFS= read -r id; do
        dunstctl history-pop "$id"
    done
```

emanuele6 (changes requested):

polybar-scripts/dunst-controller/dunstctl.sh
aidenlangley: Right, my mistake, no, hadn't advertised it in the README either, but I did leave the double-click in `dunstctl.ini`. I use the `fish` scripts personally. Thanks for the fix, will include it in the README as well.

x70b1:
What about removing the fish version?

M-Mabrouk1:
Thank for that amazing script, But I noticed that the show-all not working with me because i wasn't having the jq package, just hilight that in the README.

aidenlangley:
> What about removing the fish version?

Why?

PR #365, opened 2022-02-13
Add status-indicators-tail

xi:
I got tired of implementing a new script for every new application, so I implemented the [StatusNotifierItem](https://www.freedesktop.org/wiki/Specifications/StatusNotifierItem) spec instead. Let me know what you think!


xi:
Thanks for the feedback!

I understand that this script is on the bigger end of the spectrum (similar to `player-mpris-tail` or `info-softwarecounter`). I had thought about publishing it as a standalone project. But IMHO that would imply different kinds of guarantees that I honestly cannot make.

I feel like including the script in polybar-scripts would communicate that this is an example that people can hack on and extend to their liking. And that is exactly what I intended. I do not want to grow this to a standalone project with its own configuration system and other feature creep.

I totally understand if you do not want to merge this. I just wanted to make the case that IMHO this script is ultimately better suited for polybar-scripts than for a separate repository. Maybe that convinces you. If not I will gladly take you up on the offer to link it from the README.

x70b1:
Well if you in your repo wirte that it is just an example which needs to be changed for the own use case, then it should be working without issues.

So yeah, I would prefer this to be honest.

xi:
I can accept that. I opened https://github.com/polybar/polybar-scripts/pull/366 for the README entry.

PR #377, opened 2022-06-11
Add wakatime integration

thomasabishop:
Add output from WakaTime status bar output to Polybar


polybar-scripts/wakatime-status/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+- [jq](https://stedolan.github.io/jq/) for JSON parsing in bash
+- [openssl](https://www.openssl.org/)
+- curl 
+
+## Security 
+
+There is provision for you to store your WakaTime API key as plain text in the script file but this is not recommended. 
+
+If you wish to encrypt your password and store it separately in another file there is a [helper script](./scripts/token_encrypt.sh) that does this for you. If you choose this option the script will source your password and decrypt it before making the request to WakaTime. This isn't 100% secure but it is safer than storing you unencrypted token in the Polybar script. Obviously store your public password and the file containing the encrypted token in another non-Git directory.
```
x70b1, line 20: > This isn't 100% secure but it is safer than storing you unencrypted token in the Polybar script.

This is just wrong. You add no security here. You add complexitiy by introducing a mechanism to encrypt the token. But you put the clear text key also in a file. And probably the file is in the same directory like the script.

Its like putting the key of your door under the doormat. Keeping the door unlocked would be the same level of security.

The best way to solve this is to keep the used token in the script. Every user can decide how to change it then, like replacing the line with a password manager query that returns the token.

x70b1 (changes requested):
Thanks for your PR!

You have also a repo for this. We could also link the repo on our main README.md if you want. Then you can maintain the code in your repo.

polybar-scripts/wakatime-status/README.md
thomasabishop: Thanks for the helpful feedback. Understood. This PR removes the security options and leaves security up to the user as advised. 

Yes it would be great if you could link the source repo in the README, cheers.

thomasabishop:
@x70b1 Hi, could you please review the updated PR, thanks.

PR #378, opened 2022-06-22
Wakatime status

thomasabishop:
Hi I've made a clean pull request for this script. It takes account of previous feedback regarding unsatisfactory security options.


x70b1:
Very cool, I will check it later!

polybar-scripts/wakatime-status/polybar_wakatime.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+TOKEN="$(echo MYTOKEN | base64)"
+curl -s --header "Authorization: Basic ${TOKEN}"  https://wakatime.com/api/v1/users/current/status_bar/today  | jq -r '.data.grand_total.text';
```
x70b1, line 4: You probably have checked `curl -u` here right?
I guess it wont work, so your workaround seems to be the best solution.

polybar-scripts/wakatime-status/README.md
```diff
@@ -0,0 +1,35 @@
+# Polybar WakaTime status
```
x70b1, line 1: Can you rename the script and the directory to `info-wakatime`?
This would fit better in our directory structure.

x70b1 (changes requested):

polybar-scripts/wakatime-status/README.md
thomasabishop: Done.

thomasabishop:
Renamed dir, script and readme.

x70b1:
I made some changes:

I removed some obvious and/or little interesting lines. The `TOKEN` var should contain the pure token only. In that way you can't change the script by mistake and is easier to understand.

Some additional things:

- Should we increase the interval to something higher? 60 secs for example? The information is not that worse, but it would save some CPU cycles if you are on your laptop battery.
- Should we handle the case when there is no internet connection?
- You added the icon to you polybar module config. I would prefer it to add it to the script output.

x70b1:
My idea would something like this:

```
#!/bin/sh

TOKEN=""

wakatime_today="$(curl -sf --header "Authorization: Basic $(echo "$TOKEN" | base64)" https://wakatime.com/api/v1/users/current/status_bar/today)"

if [ -n "$wakatime_today" ]; then
    echo "#1 $(wakatime_today | jq -r '.data.grand_total.text')"
else
    echo "#2"
fi
```

Can you test it?

thomasabishop:
Thanks for your help. Your formulation generated the following error:
`./info-wakatime.sh: line 14: wakatime_today: command not found`

I have adapted it to resolve this error whilst maintaining the conditional and hash outputs. 
I have tested and works fine.

x70b1 (approved):

PR #382, opened 2022-07-25
Resize screenshots in README to equal height

totoroot:
IMO it would look way better if screenshots of the polybar modules were all of equal height.
Since standard Markdown image notation does not allow for sizing, the HTML image tag has to be used to achieve equal height for all screenshots.
The HTML tag is supported by GitHub and most user clients, so it should not cause rendering problems.

Let me know if you would like to size them to have bigger or smaller height.


totoroot:
Just for reference, this is how it would look:
image

x70b1:
Thanks for your work on this! I am a bit unsure if this a good idea. One thing is that now the link to the script is missing. It leads to the screenshot  but not to the script README.md. The other question is if the scaled screenshots are looking that good.

totoroot:
> Thanks for your work on this! I am a bit unsure if this a good idea. One thing is that now the link to the script is missing. It leads to the screenshot but not to the script README.md. The other question is if the scaled screenshots are looking that good.

Yeah, it doesn't really look that much better. I wasn't sure if it would actually be an improvement, but the script was whipped up quickly, and I thought I'd let you decide.
I'll close this PR then.

PR #343, opened 2021-08-20
Add Mullvad VPN module with extended functionality

totoroot:
There already is a script that simply displays the connection status to Mullvad VPN.

This is a script that toggles Mullvad VPN connection with `mullvad` command line client, shows the connection status and optionally the location of the VPN server currently connected to. It is possible to configure the module to cycle through available server locations on scroll too.

PR #383, opened 2022-07-27
Add reference to more advanced polybar module for Mullvad VPN

totoroot:
A while ago I opened PR #343 to extend the functionality of the polybar module for displaying the status of Mullvad VPN.
I worked on it a little more and actually separated the functionality into two modules, that allow for better customizability using the same script. I also extended the functionality a bit.

Since I now mostly use [Codeberg](https://codeberg.org) as my Git forge and mirror my repositories to GitHub once a day, I opted for maintaining it in [public repository on Codeberg](https://codeberg.org/totoroot/polybar-modules-mullvad) but people can also check it out and contribute [here on GitHub](https://github.com/totoroot/polybar-modules-mullvad).

This PR adds a reference to the GitHub repo to both the main README and the README of the rather rudimentary status module for Mullvad VPN so that people can easily find it, should they be looking for something more advanced.

I'll close the old PR in favour of this one.


x70b1:
Thanks for your work on this!
Maybe we should drop the existing script as it is very simple? We have the link to your repo so it should work for all users.

totoroot:
@x70b1 Sure, if you think so :)
I removed the basic module in another commit.

x70b1 (approved):

Issue #387, opened 2022-09-15
info-wifionice: Add French trains


mbugert:
Currently, the `info-wifionice` script only works in German ICE trains. Some trains in the French Normandie offer arrival times and speed in their train portal too, so it would be cool to integrate those in the script.

x70b1:
Fixed in 08e4f4d55cccec886b4e6874ba3b6102d282b2c3

PR #388, opened 2022-09-15
info-wifionice: Fix bug, integrate French trains, and rename the script

### LINKED ISSUE #386: info-wifionice: Train portal is reachable but script says otherwise
The VPN failsafe I added some time ago does not work as intended. The script shows "in-train portal unreachable" even in cases when it should work. :frowning_face:

mbugert:
* This PR fixes #386. The bug was that the script compared the output of `jq -e` when it should have compared the exit code.

* This PR also implements #387, with updated readme and screenshots. Right now, the script is called `info-wifionice`. With the addition of French trains, I propose that we rename it to `info-trainwifi`. I took the liberty of doing that already in this PR. If that goes a step too far, please let me know. :blush:


mbugert:
I see your point with the scripts having little overlap, but I think a single script would be more convenient for users.
Anyone user interested in train information would only need to add a single script to their polybar which "just works" across multiple trains or countries.

x70b1:
The question is if its common to use booth trains. I dont think so. For debugging reasons its easier to split them.

We have a lot of beginners here. You know how to write scripts, so if its better for you ro have one script you can just do it. 
So would prefer to split them.

x70b1:
Awesome, thanks! :sunglasses:

PR #391, opened 2022-09-18
feat(info-timew): add info-timew script

jnmcfly:
This script displays the daily tracked time. You can click it, to start tracking, using timew.

### DIFF
--- a/polybar-scripts/info-timew/README.md
+++ b/polybar-scripts/info-timew/README.md
@@ -1,13 +1,14 @@
-# Script: timew
+# Script: info-timew
 
 This script displays the daily tracked time. You can click it, to start tracking, using timew.
+
 
 ## Module
 
 ```ini
-[module/timew]
+[module/info-timew]
 type = custom/script
-exec = ~/.config/polybar/scripts/info-timew.sh
+exec = ~/polybar-scripts/info-timew.sh
 interval = 10
-click-left = "~/.config/polybar/scripts/info-timew.sh toggle"
+click-left = ~/polybar-scripts/info-timew.sh --toggle
 ```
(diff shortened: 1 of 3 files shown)


polybar-scripts/info-timew/README.md
```diff
@@ -0,0 +1,13 @@
...
+This script displays the daily tracked time. You can click it, to start tracking, using timew.
+
+## Module
+
+```ini
+[module/timew]
+type = custom/script
+exec = ~/.config/polybar/scripts/timew.sh
+interval = 2
+click-left = "if timew;then timew stop;else timew start; fi"
```
x70b1, line 12: It would be cool if this logic would be part of the script. We could use a command line argument. If its not set print the regular output.

polybar-scripts/info-timew/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+This script displays the daily tracked time. You can click it, to start tracking, using timew.
+
+## Module
+
+```ini
+[module/timew]
+type = custom/script
+exec = ~/.config/polybar/scripts/timew.sh
+interval = 2
```
x70b1, line 11: A small interval like this is always a bad idea. A script with the `tail` option would fit better. But it would probbaly things make to complicated here.

polybar-scripts/info-timew/timew.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
```
x70b1, line 1: Please rename the script also to `info-timew.sh`.

x70b1 (changes requested):
Thanks for this! Looks cool.
Just some notices and a shellcheck issue.

jnmcfly:
should be fixed :)

x70b1 (approved):

x70b1:
Very cool! :sunglasses:

PR #394, opened 2022-10-03
feat(system-bluetooth-bluetoothctl): add version support

kiabq:
Recent version of `bluetoothctl 5.65` has made it so the `paired-devices` command doesn't work. I've added a function to fix this for anyone using a newer version of bluetoothctl.


x70b1 (changes requested):
Thats cool, thanks for this!

I would not make it too complicated. Using the current command should be enough. If it doesn't work for someone, they have to fix it. 

We do that in other places as well. The  a most users here use rolling distrols like Arch, Void ..

kiabq:
I gotch'ya. I'll go ahead and close the PR. Thanks for taking a look!

PR #395, opened 2022-10-05
openweathermap-fullfeatured.sh: curl -sf missing double quotes

FromWau:


x70b1:
Why do you assume the quotes are required here?

FromWau:
> Thanks for this! Why do you assume the quotes are required here?
Hi, 
curl fails for me without the quotes.

x70b1:
This is interesting. I have no problems with it. Which shell do you use?

FromWau:
> This is interesting. I have no problems with it. Which shell do you use?

I am using zsh and apparently does zsh (and fish) treat the `?` in the URL as a wildcard. 

`zsh -c 'curl -sf https://location.services.mozilla.com/v1/geolocate?key=geoclue'`
will cause: `zsh:1: no matches found: https://location.services.mozilla.com/v1/geolocate?key=geoclue`
Escaping the `?` via `\` or by adding quotes resolves it.

x70b1:
Perfect! I merged your PR. 
Thanks for your work!

PR #398, opened 2022-10-07
Implement info-healthchecks script.

libklein:
Add a script that shows the health of services registered with [healthchecks.io](https://healthchecks.io).


polybar-scripts/info-healthchecks/README.md
```diff
@@ -0,0 +1,29 @@
+# Script: info-healthcheck
```
x70b1, line 1: Maybe 'info-healtcheck.io' is a better script name.

polybar-scripts/info-healthchecks/info-healthchecks.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+
+# Your API Key
+API_KEY="pPi37pxsjljbQRDjYoKkFtwrRikrdnZy"
```
x70b1, line 4: You key is compromised now.

x70b1 (changes requested):

polybar-scripts/info-healthchecks/info-healthchecks.sh
libklein: Rookie mistake, was using a test account though.

polybar-scripts/info-healthchecks/README.md
libklein: You mean 'info-healthchecks.io'? Other info scripts for online services do not do that either. I can see that this make sense since "healthcheck" is more general, but this would introduce "." into the directory name which may cause confusion.

x70b1: We have  [network-ipinfo.io](https://github.com/polybar/polybar-scripts/tree/master/polybar-scripts/network-ipinfo.io).
But just `info-healthcheck` would assume that this is something independent which is not the case.

libklein: Fair point. I've changed the name.

x70b1 (changes requested):

polybar-scripts/info-healthchecks.io/info-healthchecks.io.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+
+# Your API Key
```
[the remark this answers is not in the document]
libklein: I'd prefer to keep them unless you have a strong opinion on the matter. I don't tend to keep README files around, especially not in my .config such that inline documentation is helpful (to me) even if superfluous.

x70b1 (approved):

PR #399, opened 2022-10-09
Show github notifications count with proper pagination

skjnldsv:


x70b1:
What is your goal here? At the moment we don't show a counter of notifications. And I think it makes no sense to change that.

skjnldsv:
> And I think it makes no sense to change that.

Then we can close.
This shows a full counter of unread notifications.

I think it's a great addition. Feel free to close :shrug: :)

x70b1:
Alright, I close it.
But f you want to setup an own repo with your code, I would add a link to the main README.

PR #403, opened 2022-10-22
Add weatherscrapper-simple

FromWau:
## WeatherScraper

Displays weather stats.

### weather_scraper.py
Is the same as the openweathermap scripts but can display up to 48h forecasts and has much more weather info.
The `weather_scraper.py` scraps the weather.com site and returns the forecasts in json format for the given location.

### weather.sh
The `weather.sh` script uses the python script to get the next two hourly forecasts and displays the temp the icon for the weather state (sunny, foggy, ...) and the diff between the temperatures by an arrow (see example in README.md).

### DIFF
--- a/polybar-scripts/weatherscraper-simple/README.md
+++ b/polybar-scripts/weatherscraper-simple/README.md
@@ -4,12 +4,12 @@ Scraps the weather.com site and displays the weather in the bar.
 
 
 ### weather_scraper.py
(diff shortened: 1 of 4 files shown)


x70b1:
Sorry for the late response.
I would add your own repo to the link list on the main README? Than you can manage that script at your repo? How do you think about that?

FromWau:
That would be great, thanks.

PR #404, opened 2022-10-28
introduce weekday toggle to info-timew

jnmcfly:
introduce weekday toggle to info-timew to view either weekly or daily time spend


x70b1 (approved):

x70b1:
Just a hint:

You could simplify the parameters in your script with a case command. Like [here](https://github.com/polybar/polybar-scripts/blob/master/polybar-scripts/isrunning-compton/isrunning-compton.sh).

PR #406, opened 2022-12-01
* add python scripts for openwrt and huawei modems

seamus-45:
Show stats such as signal levels, sms count and traffic bandwidth.


polybar-scripts/network-huawei-modem/network-huawei-modem.py
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*- # PEP 263
+
+from huawei_lte_api.Client import Client
+from huawei_lte_api.Connection import Connection
+import huawei
+
+
+MODEM_URL = f"http://{huawei.username}:{huawei.password}@{huawei.host}/"
```
x70b1, line 9: Would it make sense to add the configuration to the same file? So that you only have to manage a single script?

polybar-scripts/network-openwrt-status/luci.py
```diff
@@ -0,0 +1,3 @@
+host = '192.168.8.1'
```
x70b1, line 1: Same as above: Would it make sense to add the configuration to the same file? So that you only have to manage a single script?

polybar-scripts/network-openwrt-status/README.md
```diff
@@ -0,0 +1,28 @@
+# Script: network-openwrt-status
+
+Print stats from OpenWrt router. Such as signal levels, sms count and traffic bandwidth.
+
+Written for Microdrive Tandem-4GT-OEM modem, but can be adopted to work with all OpenWrt routers.
+
+![screenshot](screenshots/1.png)
```
x70b1, line 7: You uploaded the same screenshot file twice. Could you check this?

x70b1 (changes requested):

polybar-scripts/network-openwrt-status/README.md
seamus-45: yes, I know. but they have the same output. I just don't use huawei modem at the moment. but if you insist, I can make an actual screenshot.

polybar-scripts/network-huawei-modem/network-huawei-modem.py
seamus-45: Sure, if you like the passwords to be leaked to github =) Separate file with 600 rights this is my whim, but you can do whatever.

polybar-scripts/network-openwrt-status/README.md
seamus-45: fixed

polybar-scripts/network-huawei-modem/network-huawei-modem.py
seamus-45: fixed!

polybar-scripts/network-openwrt-status/README.md
x70b1: Awesome.

PR #407, opened 2022-12-20
Fixed mono visualizations from not being able to appear

fuenix86:
Missing newline character prevented conf_channels from being added to temp config file properly, and Cava's default stereo visualization would always be used as a fallback instead.


x70b1:
@dzshn Any comments?

dzshn:
yeah this seems correct, do merge it

I don't use polybar anymore since I'm on wayland, I would probably be working on scripts otherwise

x70b1:
Thanks!

PR #409, opened 2022-12-31
smart implementation of openweathermap for polybar and i3status-rust

aloispichler:
Include sunset and sunrise, forecast and probability of precipitation in weather forecast.


aloispichler:
the script

x70b1 (commented):
Thanks for this PR!

PR #415, opened 2023-03-02
System-alltemps

oradkevych:
Simple script for displaying ALL hardware temperatures


x70b1:
Thanks for this PR!

PR #418, opened 2023-04-05
feat: :memo: Add notion todo module to README.md

zemmsoares:
Added a new polybar module that displays tasks from Notion with their API


x70b1 (approved):
Cool!

PR #421, opened 2023-04-14
[NewPlugin] polybar_v2ray plugin

EhsanAramide:


x70b1 (approved):

x70b1 (changes requested):
Please remove the submodule, we only set links to the user repos.

x70b1 (approved):

PR #424, opened 2023-04-24
Adjustment of the update-dnf.sh script to validate if there is a default route on the machine

alissonzuza:
so when validating that the default route does not exist , the bar will show the info No Internet instead of appearing dnf error, because it could not query the repo.


x70b1:
This looks like a complicated workaround. Also, an existing default route does not mean that internet access is available.

What about this?

```
sudo dnf updateinfo -q --list 2>/dev/null | wc -l
```

alissonzuza:
It was an adjustment I made , my fedora 38 is not listing without sudo .
Really without default route may not have internet more like , it would be in theory a machine of daily use the access would be standard.
I will reanalyze the possibility of validating the lack of internet to improve.
Thanks for the feedback

About this 2>/dev/null I put more and indifferent, I don't remember why I ended up using it I will adjust without this and check a way to correctly validate the lack of internet.
Because without internet it ends up generating a dnf error and how big the error is it disfigures the polybar.
I'm going to close this PR and do new tests

x70b1:
See the [Configuration](https://github.com/polybar/polybar-scripts/tree/master/polybar-scripts/updates-dnf#configuration) section. You have to setup `sudo`.

alissonzuza:
Tks

x70b1:
Let me know if this helps.

alissonzuza:
Help yes, I will validate and then do a PR or instead of validating the connection, I will validate if an error appears, show that it generated an error and must be checked via terminal, something like that to help anyone who is going to use it.

PR #425, opened 2023-05-21
wifi-ssid

14wual:
Displays the name of the current WiFi network in the Polybar. Updates the network name every 10 seconds and displays the result in the format defined in the Polybar configuration. It's useful for having a quick look at the name of the WiFi network you're connected to.


x70b1 (commented):
What is the difference between your script and the`%essid%` field from the [buildin network module](https://github.com/polybar/polybar/wiki/Module:-network)?

14wual:
Well, I'm dumb, and I didn't know that part of the module.

PR #426, opened 2023-06-11
Fix sh errors with POSIX read

maxdswain:
If `#!/bin/sh` is linked to dash or any strictly POSIX compliant shell, these scripts will give a `read: arg count` error as the standard `read` utility requires at least one variable's name. These changes fix this issue.


x70b1 (approved):

x70b1:
Thanks!

PR #430, opened 2023-07-12
docs: :memo: remove polybar-hackernews

zemmsoares:
Merged polybar-hackernews into polybar-news in order to make the module more modular, this way there is no need to have two scripts! sorry :100:


x70b1 (approved):
Don't worry. It's nice to see people working on their things.

PR #434, opened 2023-08-27
Fix bug in network-networkmanager script showing loopback interface

maxdswain:
Currently the script prints a loopback interface if it exists alongside WiFi, Ethernet etc. This commit fixes this behaviour.


x70b1:
Thanks!

PR #435, opened 2023-08-28
system-bluetooth-bluetoothctl: add bluetooth battery reporting

maxdswain:
This commit adds the printing of the Bluetooth device battery percentage to the script.


x70b1 (commented):
If we enable the experimental dbus settings, the battery info also appears in `bluetootctl info $device`. 

```
	Battery Percentage: 0x64 (100)
```

Maybe its  simpler to grep for that info in the output instead of changing the device mac style and use a longer dbus command?
The thing is that we grep already for the other details like name. 

On top of that we need a condition which checks if we find a value there or not as not every device has a battery.

maxdswain:
> Maybe its simpler to grep for that info in the output instead of changing the device mac style and use a longer dbus command?

Yep this is a better idea.

> On top of that we need a condition which checks if we find a value there or not as not every device has a battery.

I have added this feature and changed the script in the way suggested above.

PR #438, opened 2023-09-26
pipewire-control: add script

redxtech:
I have written a simple implementation of `pulseaudio-control` that doesn't rely on any `pulseaudio` tools. I had a hard time finding a module that didn't rely on `pactl` or `pamixer` (which I couldn't use in my script), so I wrote my own. Hopefully others find it helpful.

**Notes:**
- This does depend on a few non-standard commands, such as `rg` (ripgrep), `sd` (`sed` alternative), and `choose` (`awk` alternative). If it would be preferred, I can rewrite it to use the standard set of tools from `coreutils`.
- I included the format settings used in the screenshot, but noted them as such. If this isn't wanted, I can of course remove it.


x70b1:
Thanks for this PR!

Issue #436, opened 2023-09-06
pipewire-microphone script is broken


Crozh:
I'd be forever grateful if someone could make it so the microphone module for pipewire isn't broken. 
it is rather troublesome to check everytime on audio settings if my source is muted or not. This is fundamentally what's broken and not the muting script. 
My guess as to why is pw-cat not having the "--record --list-targets" command anymore - it prints the help menu on the bar instead. 
I can't code for shit and wanted to fix it myself, but I was unable to.

x70b1:
You could use pipewire-pulse as long as this script is broken.

Crozh:
what do you mean?

Crozh:
oops, closed the issue by mistake. 
I'm already using pipewire-pulse. 
I even tried using the pulseaudio script but it doesn't work because of pacmd being a requirement.

skjnldsv:
Fixed in https://github.com/polybar/polybar-scripts/pull/440 ?

PR #440, opened 2023-11-04
fix pipewire-microphone

allddd:
The `--list-targets` option was removed from `pw-cat` in [ecff225b](https://gitlab.freedesktop.org/pipewire/pipewire/-/commit/ecff225b11a29ecc9754e67de7a462f529e8e2a6), which broke the script (#436).
In my opinion it's not really necessary, since it can be done with `pactl`.
Other changes mainly concern style/efficiency.

### DIFF
--- a/polybar-scripts/pipewire-microphone/README.md
+++ b/polybar-scripts/pipewire-microphone/README.md
@@ -6,7 +6,6 @@ A script for showing and toggling the mute state of the PipeWire default microph
 ## Dependencies
 
 * pactl (libpulse)
-* pw-cat (pipewire)
 
 
 ## Module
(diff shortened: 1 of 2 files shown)


x70b1:
Thanks for the update!
Why is a trailing underscore required in the function names?

allddd:
@x70b1 
> Why is a trailing underscore required in the function names?

It's just a naming convention commonly used when writing shell/bash. 
It helps distinguish internal functions from external functions/tools, but one could argue that it's hardly necessary in a 40-line script.

If you do not like it, I can remove it, it will not affect functionality.

x70b1:
Its a bid odd at this place. I would also remove the vim config line.
Everything else is fine. I would pull it after these 2 small changes.

x70b1:
Thanks!

skjnldsv:
Why not `pactl get-default-source` ? :)

PR #442, opened 2024-02-01
New script: vpn-tunX-ip

bw00lley:
A simple script to show active tun adapters and their IP addresses using `ip tuntap list`, `ip addr show`, `jq` and `grep`. 

Differs from existing VPN scripts which show network names or rely on the `ifconfig` command which is becoming outdated.


x70b1:
Thanks!

PR #452, opened 2024-10-15
Change "energy" to "charge" in battery-combined-udev.sh

metam1nd:
The /sys/class/power_supply/ files seem to have been renamed to "charge_", using the script in its original form would throw division by zero errors, no values could be read under "energy_".

### DIFF
--- a/polybar-scripts/battery-combined-udev/battery-combined-udev.sh
+++ b/polybar-scripts/battery-combined-udev/battery-combined-udev.sh
@@ -15,20 +15,20 @@ battery_print() {
         ac=$(cat "$PATH_AC/online")
     fi
 
-    if [ -f "$PATH_BATTERY_0/energy_now" ]; then
-        battery_level_0=$(cat "$PATH_BATTERY_0/energy_now")
+    if [ -f "$PATH_BATTERY_0/charge_now" ]; then
+        battery_level_0=$(cat "$PATH_BATTERY_0/charge_now")
     fi
 
-    if [ -f "$PATH_BATTERY_0/energy_full" ]; then
-        battery_max_0=$(cat "$PATH_BATTERY_0/energy_full")
+    if [ -f "$PATH_BATTERY_0/charge_full" ]; then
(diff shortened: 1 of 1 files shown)


x70b1:
On which kernel?
I still have `/sys/class/power_supply/BAT0/energy_now` on `6.11.3-arch1-1`.

Maybe its different between vendors?

metam1nd:
Yeah I think you have the right idea, ChatGPT said the following:

> On most Linux systems, the battery's information can be found under /sys/class/power_supply/BAT0/, but the exact files available vary based on the kernel version and the hardware.
> 
> /sys/class/power_supply/BAT0/energy_now is typically found on systems where energy-based reporting is used (i.e., the battery reports its state in milliwatt-hours (mWh)).
> 
> /sys/class/power_supply/BAT0/charge_now is used on systems where charge-based reporting is used (i.e., the battery reports its state in milliamp-hours (mAh)).
> 
> Only one of these files (either energy_now or charge_now) is generally present, depending on how the battery hardware and driver report the battery status.

I'm using an older Dell Inspiron Laptop. So there should be a detection logic looking for either version and use the version that is present?

x70b1:
Oh, then its probably your kernel that will switch the naming in the future. Or it keeps as it is.

Somehow yes. On the other side: If your script is not working, then just change the paths. 
As this case will only be an issue at the setup a check about that sounds a bit like an overkill.

metam1nd:
> Oh, then its probably your kernel that will switch the naming in the future. Or it keeps as it is.

What do you mean? I'm on `6.11.3-arch1-1` too, same as you.

I'm gonna play around with it for a bit and add some detection logic, I'm in some kind of programming mood right now :D

I'll send you a new PR when I've got it working.

x70b1:
> What do you mean? I'm on 6.11.3-arch1-1 too, same as you.

My mistake. Have seen "I'm using an older **Debian** Inspiron Laptop.". But that was wrong.

PR #454, opened 2024-10-16
Add pihole-status

Cretezy:


Cretezy:
Updated to Pi-Hole 6. For Pi-Hole 5 and lower, use https://github.com/polybar/polybar-scripts/blob/7d514cdfa2ab4d3a8246c87a140e06e3ccce4cbe/polybar-scripts/pihole-status/pihole-status.sh

PR #459, opened 2025-04-22
info-wifionice: add support for float-typed speed

mbugert:
Deutsche Bahn has blessed us with an extra decimal point for train travel speeds.
This breaks the current script which assumes that speeds are integers.

This PR adds support for float-typed speeds.


x70b1:
Can you fix the quotes?

x70b1 (approved):

PR #466, opened 2025-10-14
Make popup calendar label be present time

jensenr30:


x70b1:
Whats the benefit of it?

jensenr30:
> Whats the benefit of it?

The current version of the calendar module doesn't display the date.  So there is nothing to click on, so there is no way to actually open the pop-up calendar.

This change fixes the module so it has a visible label.  The label can be clicked on which will open the pop up calendar.

x70b1:
The default for `label` is already the same value:
https://github.com/polybar/polybar/wiki/Module:-script#additional-formatting

jensenr30:
My mistake.  Sorry for the noise.

